PR #14, opened 2022-08-24
Update polynomial notebook with new public interface

LeikvollE:


docs/source/Polynomial.py
```diff
@@ -131,41 +133,27 @@ def ensemble_smoother():
...
         )
-        module_data.iteration_nr += 1
-
     plot_result(A_current, response_x_axis, uniform, priors, True)
 
 
 iterative_smoother()
 
-# %%
+# %% pycharm={"name": "#%%\n"}
```
eivindjahren, line 159: Looks like something unexpected extra got added here

src/iterative_ensemble_smoother/__init__.py
```diff
@@ -138,6 +139,10 @@ def update_step(
             step_length = self._get_steplength(self._module_data.iteration_nr)
         if noise is None:
             noise = np.random.rand(*S.shape)
+        if ensemble_mask is None:
```
eivindjahren, line 142: This change should probably be a separate commit and not end up in "Update polynomial notebook with new public interface".

eivindjahren (changes requested):

src/iterative_ensemble_smoother/__init__.py
eivindjahren: Done now

docs/source/Polynomial.py
eivindjahren: Let's not get too bogged down in this detail now I think.

eivindjahren (approved):
🚀

PR #15, opened 2022-08-26
Oscillator notebook

LeikvollE:


docs/source/Oscillator.py
```diff
@@ -30,25 +30,17 @@
...
 # # Example: Oscillator
 
-# %%
+# %% pycharm={"name": "#%%\n"}
 # Simple plotting of forward-model with a single response and parameters
 from matplotlib import pyplot as plt
 
-# Estimating parameters of an anharmonic oscillator
-
-# The anharnomic oscillator can be modelled by a non-linear partial differential
```
eivindjahren, line 42: This documentation has seemingly been removed, but I would think that this should be part of some markdown cell describing how the notebook works.

eivindjahren (changes requested):

eivindjahren:
Supersceded by #16

PR #31, opened 2022-09-28
Fix es-mda examples in docs

ajikmr:
Fixing es-mda examples.


ajikmr:
Hi @eivindjahren , Please merge this fix. I found these minor issues when I looked at the merged code.

eivindjahren:
As far as I understand it, this changes the example so that it used the non-iterative version of the algorithm when performing ES-MDA, is that correct?

ajikmr:
This just fixes some minor error in the previous commit. ES-MDA should iteratively call 'ies.ensemble_smoother_update_step' routine, and not 'ies.IterativeEnsembleSmoother.update_step'.

eivindjahren:
The `ies.ensemble_smoother_update_step` function keeps step-length at always 1.0, while `ies.IterativeEnsembleSmoother.update_step` will vary step_length. Which is a bit of a semantic change. Could you amend the commit message body so that it has a small description of why this needs to change?

ajikmr:
I have updated main commit message to explain changes.

eivindjahren:
> I have updated main commit message to explain changes.

I see you updated the PR message, to update the commit message body you can use the `git commit --amend` command and then use `git push -f fix-es-mda-example` to push that change to github :)

ajikmr:
Thanks @eivindjahren for your suggestion. I have updated commit message as well.

eivindjahren (approved):
:rocket:

PR #35, opened 2022-11-16
Specify in which folder tests are saved

dafeda:
This to avoid python files in, say the docs-folder being run during test-discovery.


eivindjahren:
Just missing a newline at end of file, otherwise it is fine.

eivindjahren (approved):

PR #40, opened 2022-11-21
Refactor

pinkwah:


eivindjahren (approved):
Please double check that this does not break ert and move 3.11 stuff to separate commit.

Issue #42, opened 2022-11-22
Add option to turn `A^+A` projection off


dafeda:
`A^+A` projection should only be done when the `X` in `g(X)` is of dimension `p <= N-1`, where `p` is the number of parameters and `N` the number of realizations.

In the ongoing work on implementing localization, we loop through each parameter in `X` and do a update.
We must **not** do a projection on each loop.

This means that we need to add the option to turn off projection.

PR #43, opened 2022-11-22
Add option to turn on / off projection

dafeda:
Resolves: #42 

Projection should only be done when the full global state vector has less parameters than N-1 where N are the number or realizations. The projection should not be applied when for example updating parameters one at a time, or a chunk at a time.

### DIFF
--- a/src/iterative_ensemble_smoother/experimental.py
+++ b/src/iterative_ensemble_smoother/experimental.py
@@ -36,7 +36,7 @@ def ensemble_smoother_update_step_row_scaling(
             D,
             inversion,
             truncation,
-            True,
+            False,
             np.zeros((S.shape[1], S.shape[1])),
             1.0,
             1,
--- a/src/iterative_ensemble_smoother/experimental.py
+++ b/src/iterative_ensemble_smoother/experimental.py
@@ -36,6 +36,7 @@ def ensemble_smoother_update_step_row_scaling(
             D,
             inversion,
             truncation,
+            True,
             np.zeros((S.shape[1], S.shape[1])),
             1.0,
             1,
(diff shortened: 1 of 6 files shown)


src/iterative_ensemble_smoother/experimental.py
```diff
@@ -36,6 +36,7 @@ def ensemble_smoother_update_step_row_scaling(
             D,
             inversion,
             truncation,
+            True,
```
ManInFez, line 39: should we do projection when row_scaling? I thought that was doing update one parameter at a time?

src/iterative_ensemble_smoother/ies.cpp
```diff
@@ -425,7 +425,7 @@ MatrixXd makeX(const MatrixXd &A, const MatrixXd &Y0, const MatrixXd &R,
        and when the forward model is non-linear.
        Section 2.4.3
     */
-    if (A.rows() > 0 && A.cols() > 0) {
+    if (projection && (A.rows() > 0 && A.cols() > 0)) {
```
ManInFez: Should we even be here if A has either no columns or no rows?

src/iterative_ensemble_smoother/experimental.py
dafeda: Good point, thanks.

src/iterative_ensemble_smoother/ies.cpp
dafeda: You are probably right, but I don't want to remove it in this PR.
Created an issue though:
https://github.com/equinor/iterative_ensemble_smoother/issues/44

ManInFez (approved):
LGTM

PR #50, opened 2022-11-29
Improve correctness of variable names

### LINKED ISSUE #26: Naming of Y/S response_matrix/sensitivity_matrix
Variable names S/Y or parameter named response_matrix/sensitivty_matrix needs to be reviewed that they are named correctly.

### LINKED ISSUE #28: Make sure `centered_anomaly_matrix` should in fact be `centered`
The `update_step` takes the following as a parameter:

https://github.com/equinor/iterative_ensemble_smoother/blob/6648157cadfabed1029ade23feac34a5a67e2979/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L55

As far as I can tell, this ends up being used here (only under certain conditions):

`https://github.com/equinor/iterative_ensemble_smoother/blob/6648157cadfabed1029ade23feac34a5a67e2979/src/iterative_ensemble_smoother/ies.cpp#L70`

and this function centers the `A` matrix.

I have done some preliminary testing which indicates that this should not be centred, but more analysis is needed.

dafeda:
Resolves #28 
Resolves #26

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -88,12 +88,14 @@ class IterativeEnsembleSmoother:
         :param inversion: The type of subspace inversion used in the algorithm, defaults
             to exact.
         """
-        Y = response_ensemble
-        A = parameter_ensemble
+        parameters = parameter_ensemble.shape[0]
+        realizations = parameter_ensemble.shape[1]
+        responses = response_ensemble.shape[0]
+        parameter_ensemble = parameter_ensemble
         if step_length is None:
             step_length = self._get_steplength(self._module_data.iteration_nr)
         if noise is None:
-            noise = np.random.rand(*Y.shape)
+            noise = np.random.rand(responses, realizations)
         if ensemble_mask is None:
             ensemble_mask = np.array([True] * self._ensemble_size)
         if observation_mask is None:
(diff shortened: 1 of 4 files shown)


pinkwah (approved):
I trust

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,34 +88,34 @@ def update_step(
         :param inversion: The type of subspace inversion used in the algorithm, defaults
             to exact.
         """
-        S = sensitivity_matrix
-        A = centered_anomaly_matrix
+        Y = response_ensemble
```
ManInFez, line 91: This looks a bit strange, as it is just essentially a renaming of input parameters, but I understand it then connects better to the mathematics in the code below

ManInFez (approved):
:bacon:

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: I agree that it looks a bit strange, so I've pushed a commit proposing to use longer names instead.
I've still kept the more "mathy"-names in the C++ code though.
Perhaps that's not a bad way to go about it...
Do you find this a bit easier to read @pinkwah ?

ManInFez:
should we close this and use https://github.com/equinor/iterative_ensemble_smoother/pull/51

dafeda:
> should we close this and use #51

Think I'll merge this first, then rebase and merge the other one.
It will lead to nicer "What's changed" when I create a new release.

PR #51, opened 2022-11-29
Center response ensemble before passing to C++

dafeda:
Removes the need to create centered and normalized matrix in C++,
and is more according to the notation used in the paper,
where Y referes to centered and normalized responses.

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -117,7 +117,8 @@ class IterativeEnsembleSmoother:
         update_A(
             self._module_data,
             parameter_ensemble,
-            response_ensemble,
+            (response_ensemble - response_ensemble.mean(axis=1, keepdims=True))
+            / np.sqrt(realizations - 1),
             R,
             E,
             D,
--- a/src/iterative_ensemble_smoother/_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_ensemble_smoother.py
@@ -53,7 +53,8 @@ def ensemble_smoother_update_step(
         response_ensemble = response_ensemble @ AA_projection
 
     X = make_X(
-        response_ensemble,
+        (response_ensemble - response_ensemble.mean(axis=1, keepdims=True))
+        / np.sqrt(realizations - 1),
         R,
         E,
         D,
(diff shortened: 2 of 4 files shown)


src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,34 +88,37 @@ def update_step(
...
+        if projection and (parameters < realizations - 1):
+            AA_projection = _compute_AA_projection(parameter_ensemble)
+            response_ensemble = response_ensemble @ AA_projection
 
         update_A(
             self._module_data,
-            A,
-            S,
+            parameter_ensemble,
+            (response_ensemble - response_ensemble.mean(axis=1, keepdims=True))
```
ManInFez, line 120: Does this deserve a comment?, same in ensemble_smoother

ManInFez:
also a typo in one commit message `User words with meaning as variable names` -> `Use words with meaning as variable names`

dafeda:
> I would maybe squash the first two commits (if #50 is not merged first)?

Done

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: Yeah that got merged as part of the other PR 😢

dafeda:
> also a typo in one commit message `User words with meaning as variable names` -> `Use words with meaning as variable names`

Yeah that got merged as part of the other PR 😢

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: Not sure.
I've added this description to `updateA` and `makeX` in `ies.cpp`:

```c++
/**
 * @param Y Predicted ensemble anomalies normalized by sqrt(N-1),
 *          where N is the number of realizations.
 *          See line 4 of Algorithm 1 and Eq. 30.
 */
```

These function are called from `_iterative_ensemble_smoother` and `_ensemble_smoother` which means that users of the `iterative_ensemble_smoother` package do not have to think about this centering and normalization.
Not sure if it's a good idea to duplicate the information in the docstring.
What do you think?

ManInFez (approved):
LGTM :bacon:

Issue #53, opened 2022-11-29
Remove unused `W` in `makeX´


dafeda:
https://github.com/equinor/iterative_ensemble_smoother/blob/ba3591e049baa9806efea9fcc497e2daf8fb043c/src/iterative_ensemble_smoother/ies.cpp#L436-L437

Issue #52, opened 2022-11-29
Remove unused calculation of local_costf


dafeda:
https://github.com/equinor/iterative_ensemble_smoother/blob/ba3591e049baa9806efea9fcc497e2daf8fb043c/src/iterative_ensemble_smoother/ies.cpp#L484-L490

Issue #45, opened 2022-11-23
Remove unused parameter `iteration_nr` in `makeX`


dafeda:
https://github.com/equinor/iterative_ensemble_smoother/blob/209e93f51aa8dcf4550a71dbca0b0d62a5509706/src/iterative_ensemble_smoother/ies.cpp#L408

PR #62, opened 2022-12-08
Reduce state held by `ies`

dafeda:
Resolves: #52 
Resolves: #53 
Resolves: #45 
Resolves: #63 

ERT passes matrices to `ies` that have already been sliced, i.e., deactivated observations and realizations have been removed.
One reason for `ies` to hold state was to be able to deal with observations and realizations being deactivated in-between iterations.
Assuming that users of `ies` manage this themselves allows us to hold less state and simplify the code.

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -29,7 +29,7 @@ class IterativeEnsembleSmoother:
         dec_steplength: float = 2.5,
     ):
         self.iteration_nr = 1
-        self._ensemble_size = ensemble_size
+        self._initial_ensemble_size = ensemble_size
         self.max_steplength = max_steplength
         self.min_steplength = min_steplength
         self.dec_steplength = dec_steplength
@@ -106,7 +106,7 @@ class IterativeEnsembleSmoother:
 
         num_params = parameter_ensemble.shape[0]
         num_obs = len(observation_values)
-        # Note that this may differ from self._ensemble_size,
+        # Note that this may differ from self._initial_ensemble_size,
         # as realizations may get deactivated between iterations.
         ensemble_size = parameter_ensemble.shape[1]
         if step_length is None:
@@ -122,7 +122,7 @@ class IterativeEnsembleSmoother:
         E = (E.T / observation_errors).T
         response_ensemble = (response_ensemble.T / observation_errors).T
 
-        if projection and (num_params < self._ensemble_size - 1):
+        if projection and (num_params < self._initial_ensemble_size - 1):
             AA_projection = _compute_AA_projection(parameter_ensemble)
             response_ensemble = response_ensemble @ AA_projection
 
@@ -132,7 +132,7 @@ class IterativeEnsembleSmoother:
         coefficient_matrix = update_A(
             parameter_ensemble,
             (response_ensemble - response_ensemble.mean(axis=1, keepdims=True))
-            / np.sqrt(self._ensemble_size - 1),
+            / np.sqrt(self._initial_ensemble_size - 1),
             R,
             E,
             D,
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -4,7 +4,7 @@ import numpy as np
 
 rng = np.random.default_rng()
 
-from ._ies import InversionType, ModuleData, make_D, make_E, update_A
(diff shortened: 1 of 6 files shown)


src/iterative_ensemble_smoother/ies.cpp
```diff
@@ -479,95 +336,37 @@ MatrixXd makeX(py::EigenDRef<MatrixXd> Y, py::EigenDRef<MatrixXd> R,
...
-             py::EigenDRef<MatrixXd> A, py::EigenDRef<MatrixXd> Y,
-             // Measurement error covariance matrix (not used)
-             py::EigenDRef<MatrixXd> Rin,
-             // Ensemble of observation perturbations
-             py::EigenDRef<MatrixXd> Ein,
-             // (d+E-Y) Ensemble of perturbed observations - Y
-             py::EigenDRef<MatrixXd> Din, const Inversion ies_inversion,
-             const std::variant<double, int> &truncation,
-             double ies_steplength) {
+MatrixXd updateA(Data &data,
```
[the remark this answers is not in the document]
ManInFez: maybe it could just be iteration number now?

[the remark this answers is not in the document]
dafeda: No we don't need it anymore, but I created a separate issue for it here:

https://github.com/equinor/iterative_ensemble_smoother/issues/63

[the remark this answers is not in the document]
dafeda: But perhaps it could be included here as it fits into the "reduce state" mission.
I'll add it here, thanks.

ManInFez (approved):
just a minor comment, otherwise :bacon: all the way!

src/iterative_ensemble_smoother/ies.cpp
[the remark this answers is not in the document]
dafeda: Removed

src/iterative_ensemble_smoother/ies.cpp
```diff
@@ -451,123 +293,47 @@ MatrixXd makeX(py::EigenDRef<MatrixXd> Y, py::EigenDRef<MatrixXd> R,
...
 
-  /*
-   * Allocates the local matrices to be used.
-   * Copies the initial measurement perturbations for the active observations
-   * into the current E matrix. Copies the inputs in D, Y and R into their local
-   * representations
-   */
-  MatrixXd E = data.make_activeE();
-  MatrixXd D = Din;
+  auto const ensemble_size = A.cols();
```
ManInFez, line 329: not used

tests/test_algorithm.py
```diff
@@ -88,3 +88,48 @@ def test_that_global_ies_update_is_identical_to_local():
...
+    )
+    param_ensemble = smoother.update_step(
+        response_ensemble[:, ens_mask],
+        param_ensemble,
+        obs_errors,
+        obs_values,
+        noise[:, ens_mask],
+        ensemble_mask=ens_mask,
+    )
+    param_ensemble = smoother.update_step(
```
ManInFez, line 127: should we have a realization also fail in between iterations?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -28,16 +28,12 @@ def __init__(
         min_steplength: float = 0.3,
         dec_steplength: float = 2.5,
     ):
-        self._module_data = ModuleData(ensemble_size)
+        self.iteration_nr = 1
         self._ensemble_size = ensemble_size
```
ManInFez, line 32: could maybe be renamed to `_initial_ensemble_size`

src/iterative_ensemble_smoother/ies.cpp
dafeda: Nice, thanks 👍

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: Agree 👍

tests/test_algorithm.py
dafeda: I was thinking of doing something like `ens_mask[5:] = False` between iterations to emulate failing realizations,
but that turned out just to be a test of how to correctly select columns from `param_ensemble` and not something related to the algorithm itself.
Do you still think it's useful?

ManInFez: probably not needed here. ERT should test that things work when realizations are failing

ManInFez (approved):
more :bacon:

Issue #64, opened 2022-12-10
Consider removing `R` as it is always the identity matrix


dafeda:
`R` is not needed when doing exact inversion which is what we use most of the time.
It's by default defined as `R = np.identity(len(observation_errors), dtype=np.double)` so it can take up a bit of memory if there are lots of observations.

If at some point we manage to model the covariance between observations, we would have to re-introduce it.

geirev:
It makes sense to remove R as a matrix, and then also the inversion options that use it.
If we later introduce measurement error correlations, it is better (faster and less memory-consuming) to represent them through the measurement error perturbations in the matrix E (m x N) and use that R ~ E E^T.

dafeda:
> It makes sense to remove R as a matrix, and then also the inversion options that use it. If we later introduce measurement error correlations, it is better (faster and less memory-consuming) to represent them through the measurement error perturbations in the matrix E (m x N) and use that R ~ E E^T.

That makes sense, thank you.

dafeda:
I have created a PR (https://github.com/equinor/iterative_ensemble_smoother/pull/67) that removes the possibility of using exact `R` as well as `R` estimated using low-rank `E`.
It has not really been possible to pass an exact `R` anyway and I doubt anyone has used the low rank `E` option.

We now only have exact inversion in which case we do not create `R` at all, and subspace inversion using `EE^T`.

geirev:
An afterthought is that as the iterative smoother is now a separate repository, I would suggest keeping the different functionalities (except the one with R estimated from a low-rank E), as other users might want them. Maybe it is better to disable them from the ERT menu to make them invisible to the users and not allocate space for R in ERT? After all, it is not a lot of extra code.

We will use the subspace inversion with a low-rank E next year as we are starting a project using it with ERT.

dafeda:
You write "except the one with R estimated from a low-rank E" but also that we will use subspace inversion with a low-rank E next year.
Do you mean that we should keep that one as well?
Perhaps just remove inversion with full `R` and remove the rest?

geirev:
In Equinor, "we" will use the exact inversion (0) and the inversion with a low-rank E (3).

Using a full-rank R (1) may interest other users of the inversion code, so if this is a priority, we should keep it.
Using a low-rank R computed from E (2) is useless, as using option (3) is faster and gives precisely the same result.

Thus, we can keep options (0, 1, and 3) in the iterative smoother code but deactivate option 1 in ERT.

[...]

PR #67, opened 2022-12-22
Support full observation error matrix

dafeda:
Resolves: #64

### DIFF
--- a/src/iterative_ensemble_smoother/ies.cpp
+++ b/src/iterative_ensemble_smoother/ies.cpp
@@ -153,16 +153,7 @@ void lowrankCinv(
 }
 
 /**
- * Implementation of algorithm as described in
- * "Efficient Implementation of an Iterative Ensemble Smoother for Data
- * Assimilation and Reservoir History Matching"
- * https://www.frontiersin.org/articles/10.3389/fams.2019.00047/full
- *
- * Section 2.4.3
- */
-/**
- *  The standard inversion works on the equation
- *          S'*(S*S'+R)^{-1} H           (a)
+ * Sections 3.3 and 3.4
  */
 void subspace_inversion(MatrixXd &W, const Inversion ies_inversion,
                         const MatrixXd &E,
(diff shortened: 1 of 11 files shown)


frode-aarstad (approved):
R is no more. LGTM

tests/test_theoretical_property.py
```diff
@@ -20,7 +32,7 @@ def test_that_update_is_according_to_theory():
     """
     N = 1500
     nparam = 3
-    var = 2
+    var = 2.0
```
ManInFez, line 35: why...

dafeda: Just checking that you are paying attention.
Fixed 👍

src/iterative_ensemble_smoother/ies.cpp
```diff
@@ -287,7 +274,7 @@ create_transition_matrix(py::EigenDRef<MatrixXd> Y, py::EigenDRef<MatrixXd> R,
   if (ies_inversion == Inversion::exact) {
     exact_inversion(W, S, H, ies_steplength);
   } else {
-    subspace_inversion(W, ies_inversion, E, R, S, H, truncation,
+    subspace_inversion(W, ies_inversion, E, R.value(), S, H, truncation,
```
ManInFez, line 277: Won`t `R.value()` raise a `[std::bad_optional_access](https://en.cppreference.com/w/cpp/utility/optional/bad_optional_access)` if R is not an actual matrix?

dafeda: Good question, thanks.
I've now made `R` optional in `subspace_inversion` and made some changes to how `R` is created in the first place.
I've also added tests to make sure every inversion method works when `observation_errors` are either a 1D array or a 2D array.

tests/test_theoretical_property.py
```diff
@@ -1,11 +1,32 @@
...
+var = 2.0
 
-def test_that_update_is_according_to_theory():
+
+@pytest.mark.parametrize(
+    "inversion,errors",
+    [
+        (ies.InversionType.EXACT, np.diag(np.array([var, var, var]))),
+        (ies.InversionType.EXACT, np.array([np.sqrt(var), np.sqrt(var), np.sqrt(var)])),
+        (ies.InversionType.EE_R, np.diag(np.array([var, var, var]))),
```
ManInFez: missing EE_R for np.array

ManInFez (approved):
:bacon:

Issue #65, opened 2022-12-10
Consider modeling the API after sklearn


dafeda:
See for example how sklearn does linear regression:

https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html#sklearn.linear_model.LinearRegression.fit

ES MDA would be something like this:

```python
from iterative_ensemble_smoother import ESMDA

smoother = ESMDA()

X = prior()
for weight in weights:
  smoother.fit(Y, D, weight)
  X = smoother.predict(X)
```

Have not thought this through, so needs discussions.

Issue #27, opened 2022-09-05
ensemble_smoother_update_step should be just IterativeEnsembleSmoother.update_step(step_length=1, iteration_nr=1)


eivindjahren:
`ensemble_smoother_update_step` is just a simplified version of `IterativeEnsembleSmoother.update_step` where `step_length` is always `1.0` and `iteration_nr` is always `1`. Therefore, `ensemble_smoother_update_step` should not exist, and it should be possible to perform this simplified update step by setting parameters in `IterativeEnsembleSmoother`. However, before this is done, it should be verified that these produce the same result.

dafeda:
Consider #65 as well

PR #72, opened 2023-02-10
Model API after sklearn

dafeda:
Resolves: #27
Resolves: #65

Work on localization has uncovered problems with how we deal with projection of the response ensemble when `num_params < ensemble_size - 1`.

This PR attempts to fix this issue but since it requires a breaking change in the API, I propose larger changes as well.

Edit: After proposal by @Blunde1 , the new API takes in `param_ensemble` instead of `projected_responses` whenever projection is necessary. This is nicer because it hides the function `ies.response_projection` from users.
Care must however be taken to pass in the whole `param_ensemble` and not a subset of it, i.e., the `param_ensemble` sent to `fit` and `update` will differ when doing local analysis.

## Proposed API based on `sklearn`

We have for some time discussed wanting to base the design of the iterative ensemble smoother on the design of `sklearn`.

Typical usage of `sklearn` involves calling the methods `fit` and `predict`. For example:

```python
from sklearn import linear_model
reg = linear_model.Lasso(alpha=0.1)
reg.fit([[0, 0], [1, 1]], [0, 1])

reg.predict([[1, 1]])
```

This PR proposes something similar for the `iterative_ensemble_smoother`, but instead of `predict` I propose using `update`.

Running `Ensemble Smoother`:

```python
import iterative_ensemble_smoother as ies

smoother = ies.EnsembleSmoother()
smoother.fit(
    response_ensemble,
    observation_errors,
    observation_values,
)
param_ensemble = smoother.update(param_ensemble)
```

Running `Iterative Ensemble Smoother`

```python
smoother_ies = ies.IterativeEnsembleSmoother(ensemble_size)
for _ in range(4):
    smoother_ies.fit(
        responses_ies,
        obs_errors,
        obs_values,
    )
    params_ies = smoother_ies.update(params_ies)
    responses_ies = run_forward_model(params_ies)
```

## Problem with projection of responses

Say you want to update 3 parameters and have 100 realizations.
Given that your model is non-linear, responses should be projected when doing a global update.
But what if we loop through parameters and update each one separately - should responses be projected each time?
Based on my understanding, the answer is no, the responses should not be projected in every iteration.
The reason is that the projection is basically the pseudo inverse of the parameter ensemble and we can not calculate this for each parameter and apply it to the responses.
My proposed solution is to have users decide whether to project or not.

So, in order for a global update of 3 parameters to be identical to looping through parameters, the following must be done:

```python
import iterative_ensemble_smoother as ies

if not linear and num_params < ensemble_size - 1:
    param_ensemble = get_all_parameters()

# Global update
smoother = ies.EnsembleSmoother()
smoother.fit(
    response_ensemble,
    observation_errors,
    observation_values,
    noise=noise,
    param_ensemble=param_ensemble,
)
param_ensemble = smoother.update(param_ensemble)

# Local update
for i in range(num_params):
    smoother.fit(
        Y,
        observation_errors,
        observation_values,
        noise=noise,
        param_ensemble=param_ensemble,
    )
    param_ensemble[i, :] = smoother.update(param_ensemble[i, :])
```

~~So, my solution is to pass `projected_response` to `fit` which I admit is not great, but it's the best solution I have managed to come up with.
From my understanding, `fit` has to somehow be told whether we are doing a global update or a local one, and passing `projected_response` is my way of communicating this.
The reason for passing both `response_ensemble` and `projected_response` is that `response_ensemble` is used to create `D`.~~

## ~~Projection when model is linear~~

**This seems to have been caused by the fact that I forgot to scale responses with observations inside of the `fit` function.**

~~If I am reading it correctly, `evensen2019` states that projection should not be neccessary when the model is linear as per equation (29) in the figure below.~~

projection

~~See comment below that references a test that will fail unless projection is done and where the model is linear.~~


tests/test_increased_belief_case.py
```diff
@@ -105,7 +105,11 @@ def test_that_update_for_a_linear_model(number_of_realizations):
             ]
         )
         A_posterior = ies.ensemble_smoother_update_step(
-            S, A, np.full(observations.shape, error), observations
+            S,
+            A,
+            np.full(observations.shape, error),
+            observations,
+            projected_response=S @ ies.response_projection(A),
```
dafeda, line 112: According to the paper, projection should not be necessary if the forward model is linear.
Why then is it needed here?

tests/test_properties.py
```diff
@@ -0,0 +1,383 @@
...
+
+        previous_mean_posterior = mean_posterior
+
+    # At strong beliefs, we should be close to the maximum likelihood estimate
+    assert np.all(
+        np.linalg.norm(previous_mean_posterior - maximum_likelihood) < epsilon
+    )
+
+
+def test_that_multiple_iterations_of_ies_reduces_generalized_variance():
```
dafeda, line 136: Not doing projection here causes test to fail even though model is linear.

tests/test_properties.py
```diff
@@ -0,0 +1,383 @@
...
+            noise=noise,
+            projected_response=projected_ies,
+        )
+        params_ies = smoother_ies.update(params_ies)
+        assert np.linalg.det(np.cov(prior_es)) > np.linalg.det(np.cov(params_ies))
+        # Run forward model with updated parameters
+        responses_ies = poly(obs_times).T @ params_ies
+        projected_ies = responses_ies @ ies.response_projection(params_ies)
+    # IES does not give a better fit than ES perhaps because this is a linear case
+    # assert np.linalg.det(np.cov(posterior_es)) > np.linalg.det(np.cov(params_ies))
```
dafeda, line 202: This needs to be discussed.

tests/test_properties.py
```diff
@@ -0,0 +1,383 @@
...
+            responses_ies,
+            obs_errors,
+            obs_values,
+            noise=noise,
+        )
+        params_ies = smoother_ies.update(params_ies)
+        assert np.linalg.det(np.cov(prior_es)) > np.linalg.det(np.cov(params_ies))
+        # Run forward model with updated parameters
+        responses_ies = poly(obs_times).T @ params_ies
+    # assert np.isclose(
```
dafeda, line 191: Posterior calculated using IES does not seem to tend towards posterior calculated using ES.

dafeda:
Started work on a branch of ERT in order to test the new API. (https://github.com/equinor/ert/pull/4915)
Manually ran hm-tutorial and getting minor differences.
Seems like the 14 or so first digits are identical but that there are some differences after that.
For example:

```
< "MULTFLT:MULTFLT_F1" : 0.0009606824126993932, 
> "MULTFLT:MULTFLT_F1" : 0.0009606824126988369,
```

Not sure why this is so. Perhaps some dependencies changed or something like that.
In any case, I think it's safe to update the references.

tests/test_properties.py
dafeda: This was due to the transition_matrix being stored in IterativeEnsembleSmoother and not coefficient_matrix.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -125,32 +122,88 @@ def update_step(
...
+            truncation=truncation,
+            step_length=1.0,
+            ensemble_mask=None,
+            inversion=inversion,
+            param_ensemble=param_ensemble,
+        )
+
+    def update(self, param_ensemble: npt.NDArray[np.double]) -> npt.NDArray[np.double]:
+        assert self.smoother is not None
+        return self.smoother.update(param_ensemble)
```
Blunde1, line 206: This looks correct to me. I did not remember if one iteration of SIES with step length 1.0 (as implemented here) was identical to ES, so I compared both the reference book on ES (Evensen 2009, section 9.5) with Algorithm 1 of Evensen 2019.

In the first iteration and without projection, with step-size 1.0, the update in Line 9 of Algorithm 1 can be written
`X_posterior = X_prior @ (I + Y.T @ np.linalg.inv(Y@Y.T + E@E.T ) @ H`
Which should equal Equation 9.27 of Evensen 2009.

Thus all of this should be okay! 👍

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -8,16 +8,28 @@
...
+    param_ensemble: npt.NDArray[np.double],
+) -> npt.NDArray[np.double]:
+    """A^+A projection is necessary when the parameter matrix has fewer rows than
+    columns, and when the forward model is non-linear. Section 2.4.3
+    """
+    _, _, vh = np.linalg.svd(
+        param_ensemble - param_ensemble.mean(axis=1, keepdims=True), full_matrices=False
+    )
+    projection: npt.NDArray[np.double] = vh.T @ vh
+    return projection
```
Blunde1, line 28: Is this faster than just simply writing
```suggestion
    return np.linalg.pinv(param_ensemble) @ param_ensemble
```
And also is it correct, referring to Equation 37 of Evensen 2019? Seems to be missing the block identity-zero matrix?

src/iterative_ensemble_smoother/utils.py
```diff
@@ -9,35 +9,19 @@
...
-    parameter_ensemble: npt.NDArray[np.double],
     noise: Optional[npt.NDArray[np.double]],
     observation_errors: npt.NDArray[np.double],
     observation_values: npt.NDArray[np.double],
 ) -> None:
-
     if len(response_ensemble.shape) != 2:
         raise ValueError(
             "response_ensemble must be a matrix of size (number of responses by number of realizations)"
         )
```
Blunde1, line 21: Could we optionally validate `param_ensemble`?

```suggestion
    observation_values: npt.NDArray[np.double],
    param_ensemble: Optional[npt.NDArray[np.double]],
) -> None:
    if len(response_ensemble.shape) != 2:
        raise ValueError(
            "response_ensemble must be a matrix of size (number of responses by number of realizations)"
        )
    
    if param_ensemble is not None and len(param_ensemble.shape) != 2:
        raise ValueError(
            "parameter_ensemble must be a matrix of size (number of parameters by number of realizations)"
        )
```

Blunde1 (commented):
Finished reviewing the implementation of subspace IES.
I found no errors in the implementation i think, so all changes are potentially cosmetic.
Double checked that one iteration of SIES with step-size = 1 should equal ES.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
Blunde1: But please someone else also double check this, @dafeda I think this is what we talked about, that you have checked, and it is evidently what is implemented.

Should it be commented somewhere that this is the case?

tests/test_properties.py
```diff
@@ -0,0 +1,416 @@
...
+    )
+    A_ES = smoother.update(A)
+
+    for i in range(nparam):
+        assert np.isclose(A_ES[i, :].mean(), obs_val / 2, rtol=0.15)
+
+    assert (np.abs(np.cov(A_ES) - np.identity(nparam)) < 0.15).all()
+
+
+def test_make_D():
```
Blunde1, line 291: Is this Equation 9.22 or Equation 9.24 in Evensen 2009? Could we add a comment?
From reading [the implementation](https://github.com/dafeda/iterative_ensemble_smoother/blob/9ec40e6f4b622f75d118cca637532bdee8868023/src/iterative_ensemble_smoother/ies.cpp#L292) it seems to be 9.24.

image

Blunde1 (commented):
Looked through the tests. LGTM. Added suggestion for another test and minor cosmetic things.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -53,70 +66,54 @@ def _get_steplength(self, iteration_nr: int) -> float:
...
-            `parameter_ensemble = parameter_ensemble[:, ensemble_mask]`
-        :param observation_errors: List of active measurement of errors for each observation.
-            Assumes that `observation_mask` has been applied, for example
-            `observation_errors = observation_errors[observation_mask]`
-        :param observation_values: List of active observations.
-            Assumes that `observation_mask` has been applied, for example
-            `observation_value = observation_values[observation_mask]`
-        :param noise: Optional list of noise used in the algorithm, Has same shape as
-            response matrix.
+        :param observation_errors: 1D array of measurement errors, one for each observation.
```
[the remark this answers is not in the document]
dafeda: The current implementation supports both passing a vector of standard deviations and a covariance matrix, so adding `sd` to the name might be confusing.
I find that supporting both a vector of standard deviations and covariance matrices a tad confusing, but have not been able to come up with something better. Ideas?

The docstring of `observation_errors` on the `main` branch is more descriptive in this regard and says the following:

  :param observation_errors: Array of measurement errors for each observation,
                             or covariance matrix if errors are correlated.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: My understanding of this is that since the number of rows (m) is less than the number of columns (N), the last n - m columns of V will be multiplied by zero. 
We are here calling numpy svd with `full_matrices=False` which should do the same as using the block matrix.

Here's a proof by Python :)

```python
import numpy as np
rng = np.random.default_rng()

n = 10
N = 20
X = rng.normal(size=(n, N))

A = (X - X.mean(axis=1, keepdims=True)) / np.sqrt(N - 1)

# Numpy SVD using full_matrices=False
_, _, vh_false = np.linalg.svd(A, full_matrices=False)
assert vh_false.shape == (n, N)
proj_false = vh_false.T @ vh_false

# Numpy SVD using full_matrices=True
_, _, vh_true = np.linalg.svd(A, full_matrices=True)
assert vh_true.shape == (N, N)
proj_true = vh_true.T @ vh_true

# Projection using block matrix as in evensen2019
I_block = np.zeros(shape=(N, N))
I_block[:n, :][:, :n] = np.identity(n)
proj_block = vh_true.T @ I_block @ vh_true
assert proj_block.shape == (N, N)

# Making sure projection using block matrix is the same as using numpy SVD with full_matrices=False
assert np.isclose(proj_false, proj_block).all()

# pinv gives the same result as svd approach.
# There is already a test on main that verifies this,
# but I have chosen to keep the code as is for now.
proj_pinv = np.linalg.pinv(A) @ A
assert np.isclose(proj_pinv, proj_false).all()
```

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -125,32 +122,88 @@ def update_step(
...
+        # Line 9 of Algorithm 1
+        ensemble_size = self.ensemble_mask.sum()
+        transition_matrix: npt.NDArray[np.double] = np.identity(
+            ensemble_size
+        ) + self.coefficient_matrix[self.ensemble_mask, :][
+            :, self.ensemble_mask
+        ] / np.sqrt(
+            ensemble_size - 1
+        )
+        return param_ensemble @ transition_matrix
```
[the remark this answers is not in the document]
dafeda: Yes, that reads much better 👍 
I have pushed a fix.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: I've force pushed a fix where _response_projection uses np.linalg.pinv instead.
Also, I'm now dividing by sqrt(N-1) when calculating projection, but this does not really matter it seems:
https://github.com/equinor/iterative_ensemble_smoother/issues/74

src/iterative_ensemble_smoother/utils.py
dafeda: Sure, I've pushed a commit that compares the number of columns of `param_ensemble` with `response_ensemble` and also added the check that you propose here.

tests/test_creation.py
```diff
@@ -0,0 +1,59 @@
...
+    assert isinstance(smoother_repr, ES)
+
+
+def test_that_bad_inputs_cause_nice_error_messages():
+    ensemble_size = 10
+    num_obs = 4
+    Y = np.ones(shape=(num_obs, ensemble_size))
+    obs_errors = np.ones(num_obs)
+    obs_values = np.ones(num_obs)
+    noise = np.zeros_like(Y)
```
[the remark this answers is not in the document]
dafeda: Done

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -53,70 +66,54 @@ def _get_steplength(self, iteration_nr: int) -> float:
...
 
         R, observation_errors = _create_errors(observation_errors, inversion)
 
         _validate_inputs(
             response_ensemble,
-            parameter_ensemble,
             noise,
             observation_errors,
             observation_values,
         )
```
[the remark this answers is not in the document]
dafeda: Done

tests/test_properties.py
dafeda: It is equation 9.24 yes.
We could add a comment, but perhaps it's better to create an issue where we use `H` instead of `Dprime` as `H` is used in the paper.
What do you think?

Blunde1: I think creating an issue is the better solution, given that the notation of the 2019 paper should be used (seems natural) over the 2009 book. 

I took the liberty of creating the issue and will also be happy to follow it up https://github.com/equinor/iterative_ensemble_smoother/issues/75

dafeda: That's great 👍

tests/test_properties.py
```diff
@@ -0,0 +1,416 @@
...
+
+        previous_mean_posterior = mean_posterior
+
+    # At strong beliefs, we should be close to the maximum likelihood estimate
+    assert np.all(
+        np.linalg.norm(previous_mean_posterior - maximum_likelihood) < epsilon
+    )
+
+
+def test_that_sies_converges_to_es_in_gauss_linear_case():
```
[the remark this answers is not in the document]
Blunde1: Created a separate issue https://github.com/equinor/iterative_ensemble_smoother/issues/76

Blunde1:
I will just look through the docs examples and then approve. Everything looks really good so far.

dafeda:
> I will just look through the docs examples and then approve. Everything looks really good so far.

Great, I will go through the docstrings to make sure they are updated.

Blunde1 (approved):
LGTM and approved.
Built package and ran examples in docs/source. Everything looks correct. 
I have not built the docs as my mac was fighting me. However, there are no setup-changes to the docs it seems.
Have checked tests manually and by running them. Everything looks fine.

dafeda:
> LGTM and approved. Built package and ran examples in docs/source. Everything looks correct. I have not built the docs as my mac was fighting me. However, there are no setup-changes to the docs it seems. Have checked tests manually and by running them. Everything looks fine.

Good stuff!

I got rid of the notebook from docs implementing example 5.1 from the paper as it was for testing purposes only.
We could definitely add it at a later stage.

PR #87, opened 2023-03-23
Let vectors in `E` be centered sampled from $N(0, Cdd)$

### LINKED ISSUE #85: `makeE` great again
According to `evensen2019`, $E$ is the centered measurement-perturbation matrix whose columns are sampled from $N(0, C_{dd})$ and divided by $\sqrt{N-1}$.

The current implementation does an additional scaling of each row of $E$ as a function of observation errors.
It is not clear why this is done, but the effect seems to be that the rows of the noise matrix with larger observation errors are scaled more than those with smaller observation errors.
Without this scaling, we might potentially be "unlucky" in the draw of $E$ and up with small ensemble perturbations.

Nevertheless, without having good theoretical reasons for doing any of this extra work, we should just sample from $N(0, C_{dd})$ and use that as the noise matrix.

Here's the current implementation for reference: 

```c++
MatrixXd makeE(const VectorXd &obs_errors, const MatrixXd &noise) {
  int active_obs_size = obs_errors.rows();
  int active_ens_size = noise.cols();


  MatrixXd E = noise;
  VectorXd pert_mean = E.rowwise().mean();
  E = E.colwise() - pert_mean;
  VectorXd pert_var = E.cwiseProduct(E).rowwise().sum();


  for (int i = 0; i < active_obs_size; i++) {
    double factor = obs_errors(i) * sqrt(active_ens_size / pert_var(i));
    E.row(i) *= factor;
  }


  return E;
}
```

Blunde1:
Resolves #85


Blunde1:
@dafeda Should the snapshots perhaps be expected to fail? Locally the property tests work, and only two snapshots fail, which I think is promising.
image

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -123,8 +124,24 @@ def fit(
...
+        # Columns of E should be sampled from N(0,Cdd) and centered, Evensen 2019
+        if R is not None:
+            # Assuming R and observation_errors has relation
+            # R = Cdd / np.sqrt(observation_errors) from function _create_errors()
+            Cdd = R * observation_errors**2
+        else:
+            Cdd = np.diag(observation_errors**2)
+        eval_cdd, evec_cdd = np.linalg.eigh(Cdd)
+        Cdd_sqrt = evec_cdd * np.sqrt(eval_cdd) @ evec_cdd.T
+        E = np.array([Cdd_sqrt @ z for z in noise.T]).reshape(ensemble_size, num_obs).T
```
Blunde1, line 136: could perhaps just be written `E = Cdd_sqrt @ noise` ?

dafeda: I think something like this might work:

```python
E = np.linalg.cholesky(Cdd) @ noise
```

Blunde1: Thanks, yes this is much better.

Blunde1:
The code 
```py
  # Columns of E should be sampled from N(0,Cdd) and centered, Evensen 2019
  if R is not None:
      # Assuming R and observation_errors has relation
      # R = Cdd / np.sqrt(observation_errors) from function _create_errors()
      Cdd = R * observation_errors**2
  else:
      Cdd = np.diag(observation_errors**2)
  E = np.linalg.cholesky(Cdd) @ noise
  E = E @ (
      np.identity(ensemble_size)
      - np.ones((ensemble_size, ensemble_size)) / ensemble_size
  )
```
is copied in both `ensemble_smoother_update_step_row_scaling()` and `SIES.fit()`.
Is this enough reason to wrap it in a separate utility function?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -123,8 +124,22 @@ def fit(
         if noise is None:
             noise = rng.standard_normal(size=(num_obs, ensemble_size))
 
-        E = make_E(observation_errors, noise)
+        # Columns of E should be sampled from N(0,Cdd) and centered, Evensen 2019
+        if R is not None:
+            # Assuming R and observation_errors has relation
```
dafeda, line 129: Mind explaining this in a bit more detail, as I don't quite see why it's necessary to multiply `R` by the variance of the observation_errors.

Blunde1: I thought it would be like this due to `_create_errors()`

```py
def _create_errors(
    observation_errors: npt.NDArray[np.double],
    inversion: InversionType,
) -> Tuple[npt.NDArray[np.double], npt.NDArray[np.double]]:
        R = observation_errors
        observation_errors = np.sqrt(observation_errors.diagonal())
        R = (R.T / R.diagonal()).T
    elif len(observation_errors.shape) == 1 and inversion == InversionType.EXACT_R:
        R = np.identity(len(observation_errors))
    else:
        R = None
    return R, observation_errors
```

that is called immediately in `SIES.fit()`.
The way I read it,  if `observation_errors` is passed as a covariance matrix (which is then `C_dd`, right, or am I thinking wrongly on notation?), then a scaling with the variance is performed, and the scaled covariance matrix (now `R`) and the standard deviations (root diagonal) are returned. 

If `observation_errors` is passed as a vector with standard deviations (root diagonal of `C_dd`), then `R` is taken as the identity matrix. 

In both of these cases I try to recreate `C_dd`.

dafeda: `observation_errors` is either a 1D array of standard deviatons or a 2D covariance matrix. Not the best API perhaps.
Then I think we can do something like the following:

```python
    E = np.linalg.cholesky(observation_errors) @ noise
else:
    E = np.linalg.cholesky(np.diag(observation_errors ** 2)) @ noise
``` 

May be wrong though!

Blunde1: Because we call
```py
R, observation_errors = _create_errors(observation_errors, inversion)
```
`observation_errors` is always modified. In the case when it is initially passed as a 2D covariance matrix it is modified here
```
observation_errors = np.sqrt(observation_errors.diagonal())
```
The way I read it, is it not always modified by `_create_errors` to be a 1D array of standard deviations?

dafeda: Perhaps we can make E before calling `_create_errors`?

Blunde1: Would this not require the code to be async?

Blunde1: I think I will modify `_create_errors`, and output also `E` from this function. It makes sense considering the naming as well.

Blunde1: Tried creating `E` before everything else. This requires `noise`, which requires `ensemble_size`, which must be validated first.
If not it breaks test 
`test_that_bad_inputs_cause_nice_error_messages():`

I am thinking to modify `_validate_inputs` to handle `observation_errors` as optional between 1D and 2D arrays, and place the validation first in `SIES.fit`.

Then we may create `noise`, `E`, and `R` in that order.

src/iterative_ensemble_smoother/ies.cpp
```diff
@@ -13,9 +13,15 @@ using Eigen::MatrixXd;
 using Eigen::VectorXd;
 namespace py = pybind11;
 
-enum struct Inversion { exact = 0, subspace_exact_r = 1, subspace_re = 3 };
+enum struct Inversion
```
dafeda, line 16: Are you manually re-formatting here or did you run clang-format or something like that?
If you are manually re-formatting, please run clang-format instead. 
You can do this by executing the script clang-format that's in the script folder.

I have added an issue to run clang-format as part of CI #89

Blunde1: It was automatically re-formatted. I will take a look and run the script.

dafeda (approved):
Great stuff and beautiful commit message :)

Issue #86, opened 2023-03-23
Do not transpose symmetric matrix


Blunde1:
I see
https://github.com/equinor/iterative_ensemble_smoother/blob/fe129a42e52c77c0c7ce7cc43dab13808cafbb9d/src/iterative_ensemble_smoother/utils.py#L67-L70
in the case of `observation_errors` representing covariance matrix of measurement errors.
https://github.com/equinor/iterative_ensemble_smoother/blob/fe129a42e52c77c0c7ce7cc43dab13808cafbb9d/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L88-L90
The covariance matrix is symemtric, and transposing it has no effect.

Blunde1:
Resolved by #90

PR #90, opened 2023-03-28
Scale Cdd appropriately using cov-corr transform

### LINKED ISSUE #88: Ensure that `R` is symmetric
I see
https://github.com/equinor/iterative_ensemble_smoother/blob/fe129a42e52c77c0c7ce7cc43dab13808cafbb9d/src/iterative_ensemble_smoother/utils.py#L67-L70
Here `R` will end up not being symmetric.
It is likely we wish to perform a [cov --> corr transformation](https://math.stackexchange.com/a/300775), which should correspond nicely to scaling `D` and `E` with the observation error standard deviations
https://github.com/equinor/iterative_ensemble_smoother/blob/fe129a42e52c77c0c7ce7cc43dab13808cafbb9d/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L128-L129

We would do something like
```py
R = observation_errors
observation_errors = np.sqrt(observation_errors.diagonal()) 
sd_inv = np.diag(1 / observation_errors)
R = sd_inv @ R @ sd_inv
```

Blunde1:
Resolves #88 and #86

### DIFF
--- a/src/iterative_ensemble_smoother/utils.py
+++ b/src/iterative_ensemble_smoother/utils.py
@@ -80,7 +80,7 @@ def _create_errors(
     if len(observation_errors.shape) == 2:
         R = observation_errors
         observation_errors = np.sqrt(observation_errors.diagonal())
-        R = (R.T / R.diagonal()).T
+        R = np.diag(1 / observation_errors) @ R @ np.diag(1 / observation_errors)
     elif len(observation_errors.shape) == 1 and inversion == InversionType.EXACT_R:
         R = np.identity(len(observation_errors))
     else:


src/iterative_ensemble_smoother/utils.py
```diff
@@ -80,7 +80,7 @@ def _create_errors(
     if len(observation_errors.shape) == 2:
         R = observation_errors
         observation_errors = np.sqrt(observation_errors.diagonal())
-        R = (R.T / R.diagonal()).T
+        R = np.diag(1 / observation_errors) @ R @ np.diag(1 / observation_errors)
```
dafeda, line 83: Not sure I understand this.
What will the dimension of `R` be after transformation if for example `observation_errors` is a 2x2 matrix?
In this case, I think `R` will end up just being a scalar, but I may very well be wrong.

Blunde1: The dimensions should be the same as the input covariance matrix, I believe, given that all matrices involved are square with the same dimension.

Let $\Sigma\in R^{n\times n}$, and  $\Lambda = \text{diag}(\sigma)\in R^{n\times n}$ the diagonal matrix with standard deviations along its diagonal, then the correlation matrix is given by $\Lambda^{-1}\Sigma\Lambda^{-1}\in R^{n\times n}$.
`np.diag(1 / observation_errors)` is the inverse of the diagonal matrix with standard deviations.

The correlation matrix has elements $Cov(x_i, x_j)/(\sigma_i \sigma_j)$ which is what will be obtained when scaling the noise or error matrices, like $E$, with its standard deviations. This because we obtain a covariance matrix through $EE^T$, where the transpose makes sure $\sigma_i \sigma_j$ arrives in every denominator.

Why we seek to scale the different matrices as such is beyond me (but I think numerical stability?) since I have not yet read the inversion code yet, but this transform should correspond to how we scale `E` and `D` with the standard deviations in `observation_errors`.

Blunde1: A little sketch creating a sample covariance matrix
image

Blunde1: But `observation_errors` has been modified to a 1D array of the standard deviations, previous to the scaling of `R`
```
observation_errors = np.sqrt(observation_errors.diagonal())
```
So I still think everything should be okay.

Blunde1: Copied over the code exactly here and print `R` and its dimensions
image

dafeda: > But `observation_errors` has been modified to a 1D array of the standard deviations, previous to the scaling of `R`
> 
> ```
> observation_errors = np.sqrt(observation_errors.diagonal())
> ```
> 
> So I still think everything should be okay.

Ah, you are right, I did not read the that line 👍

dafeda (approved):
Nice 👍

PR #91, opened 2023-03-29
Projection property test

### LINKED ISSUE #77: SIES projection property test
The effect of the difference between Equations 28 and 29
image
It is hard to deterministically check that one sample is closer to some distribution than some other sample, we may however view this as an optimization problem 
image
and evaluate the ensemble of cost functions for the solutions with and without projection. It is then natural to assume that using projection when necessary should provide a better solution.

It seems natural to test this using the non-linear scalar example `g(x)=x+beta*x**3` of Section 5.1 in Evensen 2019.

Blunde1:
Resolves #77


Blunde1:
For some reason, the property test using `ies` is flaky. Note that it is not flaky using a bespoke implementation of the algorithm in Evensen 2019, as shown in the notebook.

tests/test_properties.py
```diff
@@ -40,6 +40,108 @@ def eval(self, x):
...
 
+def g(x):
+    """Non-linear model, Example 5.1 from Evensen 2019"""
+    x1 = x[0]
+    return x1 + 0.2 * x1**3
+
+
+@pytest.mark.parametrize("N", [100, 200])
+def test_that_projection_is_better_for_nonlinear_forward_model_big_N_small_p(N):
+    # For non-linear forward model, g, and m<N,
```
dafeda, line 51: Minor minor but perhaps user `p` in `m<N`.

tests/test_properties.py
```diff
@@ -40,6 +40,108 @@ def eval(self, x):
...
+    # Eq.27 should provide a better update than Eq. 28.
+    # "Better" in terms of a better optimum, given by loss in Eq. 10
+    # We here solve using both Equations 27 and 28 (with/without projection)
+    # and then evaluate the loss function.
+
+    def loss_function(xj, xj_prior, dj, Cxx, Cdd, g):
```
dafeda, line 57: Just a question, but do you think we would get the same results using Eq. 18 as loss function instead of Eq. 10?

tests/test_properties.py
```diff
@@ -40,6 +40,108 @@ def eval(self, x):
...
+        )
+
+    N = 100
+    m = 1
+    x_true = np.array([-1.0])
+
+    # sample parameters from prior
+    x_sd = 1.0
+    prior_bias = 0.5
+    X_prior = np.array(
```
dafeda, line 71: Don't think that `np.array` nor the `reshape` is necessary as `np.random.normal` with `size=(m, N)` should return an array of size `(m, N)`.

Also, might as well use `rng = np.random.default_rng()` as random number generator as it is recommended by numpy.

tests/test_properties.py
Blunde1: Yes, I think so. The loss in Eq. 18 is equivalent to Eq. 10 but constrained to searching for a solution in the ensemble subspace through only considering updates of the form Eq. 16. Since the solutions (with and without projection) both are of the form Eq. 16, evaluating them with Eq. 10 should be equivalent to directly evaluating Eq. 18.

tests/test_properties.py
Blunde1: Thanks. I see Evensen 2019 employs `m`, so I guess I should change the naming of the test. 
image

tests/test_properties.py
Blunde1: Thanks, I will try this. 👍

tests/test_properties.py
dafeda: Evensen uses `m` for number of responses / observations I think. He uses `n` for number of parameters.

Blunde1: You speak the truth, @dafeda

Blunde1: Should be okay now
1632919
it now separates between `m` and `n`. 
Since both equals 1 in this special case, I went a bit fast and used `m` for `n`.

tests/test_properties.py
```diff
@@ -40,6 +40,105 @@ def eval(self, x):
...
+
+    # Evaluate response ensemble
+    gX = np.array([g(parvec) for parvec in X_prior.T]).reshape(m, N)
+
+    # define observations
+    d_sd = np.array([1.0])
+    d = np.array([g(x_true) + rng.normal(0.0, d_sd)])
+
+    # define noise to perturb observations
+    Cdd = np.diag([d_sd**2]).reshape(m, m)
```
dafeda, line 82: Reshape necessary I think.

dafeda (approved):
Nice 👍

tests/test_properties.py
Blunde1: Yes, when m=1, otherwise it is automatic.

Issue #99, opened 2023-05-11
Implement identity matrix addition trick


tommyod:
https://github.com/equinor/iterative_ensemble_smoother/blob/5e184239924c21822ee5232c1d1544b356359c4b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#LL183C12-L183C12

If we write this as 
```
W.flat[::W.shape[0] + 1] += 1
```
we don't have to add the zeroes.

![image](https://github.com/equinor/iterative_ensemble_smoother/assets/10076072/e83731c2-ccc5-4229-9050-e9b3a63d68b0)

PR #103, opened 2023-05-12
Avoid forming identity matrix, avoid np.outer in indexing

tommyod:
Closes https://github.com/equinor/iterative_ensemble_smoother/issues/99

- Use `np.idx_` for indexing, instead of (1) indexing first on rows, then on cols and (2) instead of forming N x N matrix with `np.outer`
- Flatten array and add 1 on every (shape + 1) element instead of explicitly forming identity matrix and adding. Saves both memory usage and computational time

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -150,7 +150,7 @@ class SIES:
             response_ensemble = (response_ensemble.T / observation_errors).T
 
         if ensemble_mask is None:
-            ensemble_mask = np.array([True] * ensemble_size)
+            ensemble_mask = np.ones(ensemble_size, dtype=bool)
 
         self.ensemble_mask = ensemble_mask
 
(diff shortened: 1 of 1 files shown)


tommyod:
Take a look when it suits you @dafeda . This PR is not very meaningful, but I think it's an improvement and it's a way to start reading and getting comfortable with the code.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -150,7 +150,7 @@ def fit(
             response_ensemble = (response_ensemble.T / observation_errors).T
 
         if ensemble_mask is None:
-            ensemble_mask = np.array([True] * ensemble_size)
+            ensemble_mask = np.ones(ensemble_size, dtype=bool)
```
dafeda, line 153: This is neat, I did not think of using `np.ones` with `dtype=bool` 👍

dafeda (approved):
These are some real nice tricks that I'm sure we can use in other places in the code-base.
Well done 👍

PR #106, opened 2023-05-31
Reproducibility, linalg speedup, renaming

tommyod:
- Using `random_state` as input along with a `check_random_state()` function to facilitate reproducibility. This is how sklearn and other packages typically do it.
- Replaced the `len(array.shape)` checks with `array.ndim`. Hopefully a bit more readable
- Renamed a variable
- Avoid forming two diagonal matrices to pre-and-post multiply, instead just multiply each row/column. Goes from `O(m^3)` to `O(m^2)`. Probably not a big practical speedup, since this is unlikely to be a bottleneck, but might as well optimize these kinds of operations in my opinion.

This is a first step toward getting rid of `noise`, which I will do in another PR

I chose to just pass an integer `seed` instead of allowing a random state object, after reading
- https://github.com/scikit-learn/scikit-learn/issues/14042
- https://github.com/scikit-learn/scikit-learn/issues/16988
- https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng

I don't see a usecase where an integer seed is not good enough.

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -10,7 +10,6 @@ from ._ies import InversionType, create_coefficient_matrix
 from iterative_ensemble_smoother.utils import (
     _validate_inputs,
     _create_errors,
-    check_random_state,
 )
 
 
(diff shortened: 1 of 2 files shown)


dafeda (approved):
Very nice 👍

PR #108, opened 2023-06-02
Use seeds, steplength function, generate from multivariate normal

### LINKED ISSUE #102: Reproducibility: use random_state like sklearn
https://github.com/equinor/iterative_ensemble_smoother/blob/5e184239924c21822ee5232c1d1544b356359c4b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L9

We can consider using a `random_state` parameter in conjunction with the `check_random_state()` function (can copy this from sklearn, no reason to add dependency for these lines of code) to allow users to control random number generation for reproducibility.

https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/linear_model/_stochastic_gradient.py#L412

https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html

### LINKED ISSUE #104: Proposition: remove `noise` parameter
@dafeda Thoughts on this?

I propose removing the noise parameter altogether and writing the lines following

https://github.com/equinor/iterative_ensemble_smoother/blob/2bc4bb514058bfd6ec5c4ff13b21056227229fb8/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L123

as something like:

```python3
# A covariance matrix was passed
# Columns of E should be sampled from N(0,Cdd) and centered, Evensen 2019
    E = rng.multivariate_normal(mean=np.zeros_like(num_obs), cov=observation_errors, size=ensemble_size).T
# A vector of standard deviations was passed
else:
    E = rng.normal(loc=0, scale=observation_errors, size=(ensemble_size, num_obs)).T
    
assert E.shape == (num_obs, ensemble_size)

E -= E.mean(axis=1, keepdims=True)
```

I think this is an improvement since:
- `noise` is a superfluous input parameter, since it's only used in linear combinations with `observation_errors`. Double one means halving the other, and you get the same results. So fixating one (noise) by removing it does not decrease the number of possible inputs.
- The fact that we use cholesky factorizations is an implementation detail. What we really want is to sample from a multivariate normal, and numpy has code for this. Better to delegate to numpy than to implement our own sampler.
- One parameter less to deal with when creating an instance. Randomness is better controlled by `rng` than by `rng` in combination with a random vector as input.

I might've missed something that makes removing `noise` a bad idea. Let me know!

### LINKED ISSUE #107: Refactor _get_steplength
Instead of having 3 arguments go into `SIES` that determine the step length from an exponential decay given by `_get_steplength`, I propose that the user supplies a callback. This is the approach taken in e.g. [keras](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/LearningRateScheduler). The advantage is:

- More general approach, if a user wants to try a different schedule, it is more easily done.
- Hides implementation details from the user, the 3 variables max_steplength, min_steplength, dec_steplength are not really important for most users. Nor do they pertain to the SIES itself, they are only related to one particular part of it.

In pseudocode, I propose:

```python
def exponential_decay(iteration, learning_rate)
    # a schedule function takes an iteration index (integer, indexed from 0) and 
    # current learning rate (float) as inputs and returns a new learning rate as output (float).
    # logic here
    return learning_rate


class SIES:

    def __init__(
        self,
        ensemble_size: int,
        *,
        schedule: Union[callable, None],
    ):
    if schedule is None:
        self.schedule = exponential_decay
    else:
        self.schedule  = schedule 

```

tommyod:
- Change from relative imports to imports using the package name. This means that Python files in the package can be run individually, it also helps pytest find doctests.
- Removed the noise parameter. Reproducibility is now done via a `seed`.
- Added an argument `steplength_schedule` with signature `steplength_schedule(iteration) -> step_length`. This function deals with all issues related to step length. See issue #107 
- Use random number generator to sample from multivariate normal, not cholesky directly (but using the fast `cholesky` method)

Closes #104 
Closes #102 
Closes #107

## Warning: breaking change

- The API changes
- The random generation changes, even with the same seed, because of how rows/cols are generated

**Before** (this matrix would be multiplied with an identity matrix if we assume independent standard normals)
```
In [20]: rng = np.random.default_rng(1)

In [22]: rng.standard_normal(size=(3, 2)).T
Out[22]:
array([[ 0.34558419,  0.33043708,  0.90535587],
       [ 0.82161814, -1.30315723,  0.44637457]])
```

**After**
```
In [16]: rng = np.random.default_rng(1)

In [17]: rng.multivariate_normal(np.zeros(3), cov=np.eye(3), method="cholesky", size=2)
Out[17]:
array([[ 0.34558419,  0.82161814,  0.33043708],
       [-1.30315723,  0.90535587,  0.44637457]])
```

Notice how the matrices in both examples contain the name numbers, but they are permuted. This means we have to change the snapshots, even when the seed is identical.


tommyod:
For the life of me I could not get doctests to work on GH actions with the following setup:

```text
[testenv:py{38,39,310,311}]
deps =
    .[dev]
commands =
    pytest tests/
    pytest src/iterative_ensemble_smoother --doctest-modules
```

Of course it all worked locally by running `tox`. I give up on doctests for now.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -199,32 +178,33 @@ def update(self, param_ensemble: npt.NDArray[np.double]) -> npt.NDArray[np.doubl
     def __repr__(self) -> str:
         return (
             f"SIES(ensemble_size={self._initial_ensemble_size}, "
-            f"max_steplength={self.max_steplength}, min_steplength={self.min_steplength}, "
-            f"dec_steplength={self.dec_steplength})"
+            f"steplength_schedule={self.steplength_schedule}, seed={self.seed})"
```
dafeda, line 181: Should we have optional input parameters in __repr__?

tests/snapshots/test_snapshots/test_ensemble_smoother_update_step/test_ensemble_smoother_update_step.csv
```diff
@@ -1,3 +1,3 @@
-1.967146767000000018e+00,1.887802729999999984e+00,1.521310943000000027e+00,5.066928569999999965e-01,1.386576888000000007e+00,1.430903387000000082e+00,8.768934029999999602e-01,1.026742226999999952e+00,1.841859796000000049e+00,2.096290783999999796e+00
-5.100308150000000262e-01,-7.912242240000000315e-01,1.429665179999999869e-01,3.900854710000000170e-01,-7.375762200000000046e-01,-8.464409999999999989e-02,-1.366024670000000052e-01,-3.888128559999999845e-01,-4.576220990000000044e-01,7.570495100000000643e-02
--1.943097159999999934e-01,5.865930440000000079e-01,-9.175710600000000494e-02,3.380177240000000194e-01,1.868880370000000068e-01,-8.676134999999999750e-03,-4.114721799999999929e-02,2.734158270000000002e-01,5.822034710000000279e-01,-2.447357490000000024e-01
+1.571474030999999938e+00,2.053739563000000157e+00,1.527342551999999909e+00,1.361049552999999968e+00,1.466300458999999945e+00,1.053619995999999892e+00,1.387315756000000011e+00,8.378489999999999549e-01,1.283992301999999919e+00,1.999536568999999986e+00
```
dafeda, line 1: Have not read through the changes carefully yet.
What causes the need to update snapshots?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: Good question. If we want the `eval(repr(instance)) == instance` property to hold, then we do. But it will not hold when we pass in a function anyway. So maybe wanting this property to hold is overkill.

tests/snapshots/test_snapshots/test_ensemble_smoother_update_step/test_ensemble_smoother_update_step.csv
tommyod: Just the fact that using `rng` and sampling from normal/MVnormal was not identical to generating `noise`. Functionally it should be the same

dafeda: Am I understanding correctly that using `multivariate_normal` with method set to "cholesky" is different from doing it ourselves?

tests/test_properties.py
```diff
@@ -370,6 +369,7 @@ def test_that_update_correctly_multiples_gaussians(inversion, errors):
     obs_val = 10
     observation_values = np.array([obs_val, obs_val, obs_val])
     smoother = ies.SIES(N)
+    print(errors)
```
dafeda, line 372: Don't think we need this.

tests/snapshots/test_snapshots/test_ensemble_smoother_update_step/test_ensemble_smoother_update_step.csv
tommyod: Yes, but only different in the sense that it's like using a different seed.

tommyod:
Updated PR description with an comment on why snapshots had to be re-generated.

We should probably update version. And maybe work on the explanation for why results differ.

dafeda (approved):
Good stuff 👍

PR #113, opened 2023-06-23
Rewrite cpp to Python

### LINKED ISSUE #101: Replace `InversionType` with string
It's more pythonic to for example do ES(inversion="exact").

https://github.com/equinor/iterative_ensemble_smoother/blob/5e184239924c21822ee5232c1d1544b356359c4b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L206

### LINKED ISSUE #112: Rewrite cpp Eigen code to python NumPy
Is there any reason not to write [ies.cpp](https://github.com/equinor/iterative_ensemble_smoother/blob/main/src/iterative_ensemble_smoother/ies.cpp) as numpy code?

I can think of two valid reasons:

- Eigen is faster
- There are operations (e.g. loops or linalg operations) that are hard to express in numpy

Looking at the code, it looks fairly straightforward. Should I translate it?

This would get rid of the c++ toolchain/requirement, and perhaps simplify the code and debugging a bit.

tommyod:
**Changes**

- Rewrote every C++ linear algebra routine from Eigen to numpy/scipy
- Introduced scipy as a dependency
- Removed everything in build that was associated with C++ files
- Removed platform specific wheels (delegated to numpy/scipy wheels)
- An order of magnitude faster on diagonal covariances, same speed for dense covariances

Closes #112 
Closes #101 

**Some further comments**

- I more or less did a direct translation, but added some comments
- I removed mypy in `ies.py` since I intend to do more rewrites. I support using it in the entire codebase, but I don't want to add it right now since I hope to refactor heavily in the future

**Speed**

I measured the speed in the main branch (C++) vs this branch using this code:

```python3
def test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000):
    param_ensemble = np.random.normal(size=(num_params, ensemble_size))
    Y = np.ones(shape=(num_obs, ensemble_size))
    obs_errors = np.ones(num_obs)
    obs_values = np.ones(num_obs)
    
    from time import perf_counter
    st = perf_counter()
    model = SIES(ensemble_size)
    model.fit(Y, obs_errors, obs_values)
    model.update(param_ensemble)
    print(f"Ran in {round(perf_counter() - st, 4)} s")
```

3 runs with this code:
```text
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 10.5245 s
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 9.9963 s
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 11.1507 s
```

3 runs with the old code (C++, main branch):
```text
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 80.4747 s
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 80.0047 s
test_timing(ensemble_size=1000, num_obs=100_000, num_params=10_000)
Ran in 80.5045 s
```

Changing to a dense covariance:
```python3
E = np.random.randn(num_obs, num_obs)
obs_errors = E.T @ E
obs_values = np.ones(num_obs)
```

3 runs with this code:
```text
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 5.9958 s
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 6.3646 s
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 5.7253 s
```

3 runs with the old code (C++, main branch):
```text
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 6.1468 s
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 5.8837 s
test_timing(ensemble_size=100, num_obs=10_000, num_params=1_000)
Ran in 5.5138 s
```


tommyod:
`test_that_nans_produced_due_to_outliers_in_responses_are_handled``is flaky. The code raises `ValueError` about NaNs on some systems, but not on others. Still thinking about this test. Apart from the test the PR is good to go.

src/iterative_ensemble_smoother/__init__.py
Blunde1, line 1: When removing `InversionType` why is not a corresponding enum in Python defined?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -142,7 +143,7 @@ def fit(
 
         self.ensemble_mask = ensemble_mask
 
-        W: npt.NDArray[np.double] = create_coefficient_matrix(
+        W: npt.NDArray[np.double] = create_coefficient_matrix(  # type: ignore
```
Blunde1, line 146: Why are we ignoring? Because new `ies` Python module is not annotated?

src/iterative_ensemble_smoother/ies.py
Blunde1, line 1: Could we add type hints that are checked?

Blunde1 (commented):
I will need to go through the changes to `upload_to_pypi.yml` with you @tommyod. Here I do not have any experience, so I would like an explanation that makes sense before approval.

Removal of c++ files and references to c++ in README and setup etc. are obviously okay, given that new `ies` Python module does what it should.

The new module immediately looks okay. Will go through it more in detail now.

Other than that only a few comments for now.

Blunde1:
> `test_that_nans_produced_due_to_outliers_in_responses_are_handled` is flaky. The code raises `ValueError` about NaNs on some operating systems, but not on others. Still thinking about this test. Apart from the test the PR is good to go.

@tommyod is this flakyness new? You seem to only have removed code here, or am I missing something?

tommyod:
image

Even more flaky - it can fail or pass on different python versions.
Same version of numpy and scipy.

src/iterative_ensemble_smoother/__init__.py
tommyod: Convention mainly. I don't think I've ever seen a scientific Python package use enums, also see https://github.com/equinor/iterative_ensemble_smoother/issues/101

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: Yes

src/iterative_ensemble_smoother/ies.py
tommyod: Yes, I want to to do it, but in a future PR. I plan to rewrite the code further in the future, but in this PR I only want to rewrite from C++ to python, and do as little as possible if it's not related to the rewrite.

tommyod:
TODO: look into memory usage. maybe @JHolba can help?

.github/workflows/upload_to_pypi.yml
Blunde1, line 1: If I understand correctly, this simplification is now possible because it is not pure Python.
We may then only build a universal wheel file, and not specific platform wheel 
https://realpython.com/python-wheels/

Blunde1 (commented):
Everything exact the main `ies` LGTM.
The `ies` module I need to go through more, preferably with @JHolba to test memory usage.
open question to scout team: considering we will refactor the code more, should we drop adding type-hints (for now)?

Blunde1:
Note to self: 
- when going through `ies.py`, all functions should have the same names as in `ies.cpp`, I can test side-by-side
- If I _do_ find something, it means I need to write additional tests. Currently tests are passing, and we believe everything should be fine.

src/iterative_ensemble_smoother/ies.py
```diff
@@ -0,0 +1,385 @@
...
+    # Equation (51) states that
+    # W_{i+1} = W_i - step * (W_i - (S.T @ S + I)^-1 @ S.T @ H)
+    # Here we form
+    # C = S.T @ S + I
+    # This is a square symmetric matrix, instead of taking the SVD we
+    # call sp.linalg.solve. This is equally fast.
+
+    # Form the matrix C
+    _, ensemble_size = S.shape
+    C = S.T @ S  # TODO: Only form upper part of this matrix
```
Blunde1, line 149: It is done in the same way in the C++ code, so okay in this PR. The todo should be create issue to only form upper part of this matrix.

Blunde1 (approved):
Reviewed `ies.py` lastly.
All exposed c++ functions in `ies.cpp` exists in `ies.py` and importantly do the same thing.
Code is simplified and overall documentation quality is increased. Comments on equations and references to ies paper (evensen 2019) and DA book by Evensen (evensen 2009) checks out.
Memory tests pass.
Speed tests has been checked and are good.
I was unsure type hints should be added to `ies.py` but since it will all be refactored, and the code is not too complex, and is tested thoroughly, it should be okay.
LGTM!
Great job!

PR #123, opened 2023-08-07
Refactor IES

### LINKED ISSUE #115: Don't sample in every fit call
Currently we sample in every call to fit, but the random number generator is the same. This means the result of sampling is the same in every iteration, which is correct.

However, it is slow. We should sample once and use the same draws in every fit call.

https://github.com/equinor/iterative_ensemble_smoother/blob/24d771149134299102f8770c397b670c1d1efbf8/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L101

tommyod:
Some houesekeeping on IES:

- Renamed `_create_errors` to `covariance_to_correlation`. Wrote a docstring for it. Wrote some doctests for it. Removed redundant arg `inversion`
- Removed generation of `E` (samples from covariance) to it's own method. This closes #115
- Wrote a long comment about rescaling the covariance matrix to a correlation matrix, since there was no comments regarding this and it's not mentioned in the paper
- Renamed `iteration_nr` to `iteration`

Some of these changes are a bit subjective, so I am open to discussion.

### DIFF
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -35,7 +35,7 @@ class SIES:
         seed: Optional[int] = None,
     ):
         self._initial_ensemble_size = ensemble_size
-        self.iteration_nr = 1
+        self.iteration = 1
         self.steplength_schedule = steplength_schedule
         self.coefficient_matrix = np.zeros(shape=(ensemble_size, ensemble_size))
         self.rng = np.random.default_rng(seed)
@@ -113,6 +113,10 @@ class SIES:
             less than ensemble_size - 1 and the dynamical model is non-linear.
         """
 
+        # ---------------------------------------------------------------------
+        # ----------------- Input validation and setup ------------------------
+        # ---------------------------------------------------------------------
+
         _validate_inputs(
             response_ensemble,
             observation_errors,
@@ -126,11 +130,20 @@ class SIES:
         # Determine the step length
         if step_length is None:
             if self.steplength_schedule is None:
-                step_length = steplength_exponential(self.iteration_nr)
+                step_length = steplength_exponential(self.iteration)
             else:
-                step_length = self.steplength_schedule(self.iteration_nr)
+                step_length = self.steplength_schedule(self.iteration)
 
         assert 0 < step_length <= 1, "Step length must be in (0, 1]"
+
+        if ensemble_mask is None:
+            ensemble_mask = np.ones(ensemble_size, dtype=bool)
+
+        self.ensemble_mask = ensemble_mask
+
+        # ---------------------------------------------------------------------
+        # ----------- Computations corresponding to algorithm 1 ---------------
+        # ---------------------------------------------------------------------
 
         # Draw samples from N(0, C_dd)
         E = self._get_E(
@@ -182,16 +195,11 @@ class SIES:
         _response_ensemble -= _response_ensemble.mean(axis=1, keepdims=True)
         _response_ensemble /= np.sqrt(ensemble_size - 1)
 
-        if ensemble_mask is None:
-            ensemble_mask = np.ones(ensemble_size, dtype=bool)
-
-        self.ensemble_mask = ensemble_mask
-
         W: npt.NDArray[np.double] = create_coefficient_matrix(  # type: ignore
             _response_ensemble,
             R,  # Correlation matrix or None (if 1D array was passed)
             E,  # Samples from multivariate normal
-            D,
+            D,  # D - G(x) in line 7 in algorithm 1
             inversion,
             truncation,
             self.coefficient_matrix[np.ix_(ensemble_mask, ensemble_mask)],
@@ -203,7 +211,7 @@ class SIES:
                 "Fit produces NaNs. Check your response matrix for outliers or use an inversion type with truncation."
             )
 
-        self.iteration_nr += 1
+        self.iteration += 1
 
         # Put the values back into the coefficient matrix
         self.coefficient_matrix[np.ix_(ensemble_mask, ensemble_mask)] = W
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -8,7 +8,7 @@ if TYPE_CHECKING:
 
 from iterative_ensemble_smoother.utils import (
     _validate_inputs,
-    _create_errors,
+    covariance_to_correlation,
     steplength_exponential,
     response_projection,
 )
@@ -116,7 +116,7 @@ class SIES:
 
         E -= E.mean(axis=1, keepdims=True)
 
-        R, observation_errors_std = _create_errors(observation_errors, inversion)
+        R, observation_errors_std = covariance_to_correlation(observation_errors)
 
         # Store D as defined by Equation (14) in Evensen (2019)
         self.D_ = E + observation_values.reshape(num_obs, 1)
@@ -145,8 +145,8 @@ class SIES:
 
         W: npt.NDArray[np.double] = create_coefficient_matrix(  # type: ignore
             _response_ensemble,
-            R,
-            E,
+            R,  # Correlation matrix or None (if 1D array was passed)
+            E,  # Samples from multivariate normal
             D,
             inversion,
             truncation,
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -40,7 +40,13 @@ class SIES:
         self.coefficient_matrix = np.zeros(shape=(ensemble_size, ensemble_size))
         self.rng = np.random.default_rng(seed)
 
-    def _get_E(self, *, observation_errors, observation_values, ensemble_size):
+    def _get_E(
+        self,
+        *,
+        observation_errors: npt.NDArray[np.double],
+        observation_values: npt.NDArray[np.double],
+        ensemble_size: int,
+    ) -> npt.NDArray[np.double]:
         """Draw samples from N(0, Cdd). Use cached values if already drawn."""
 
         # Return cached values if they exist
@@ -67,7 +73,7 @@ class SIES:
         # Center values, removing one degree of freedom
         E -= E.mean(axis=1, keepdims=True)
 
-        self.E_ = E
+        self.E_: npt.NDArray[np.double] = E
         return self.E_
 
     def fit(
--- a/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
+++ b/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
@@ -135,7 +135,6 @@ class SIES:
 
         if ensemble_mask is None:
             ensemble_mask = np.ones(ensemble_size, dtype=bool)
-
         self.ensemble_mask = ensemble_mask
 
         # If it's the first time the method is called, create coeff matrix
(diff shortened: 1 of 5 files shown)


src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -75,6 +113,10 @@ def fit(
             less than ensemble_size - 1 and the dynamical model is non-linear.
         """
 
+        # ---------------------------------------------------------------------
```
dafeda, line 116: Not a big fan of these large headers.
You like?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,66 +130,76 @@ def fit(
...
+        # Transform covariance matrix to correlation matrix
+        R, observation_errors_std = covariance_to_correlation(observation_errors)
 
         # Store D as defined by Equation (14) in Evensen (2019)
         self.D_ = E + observation_values.reshape(num_obs, 1)
         D = self.D_ - response_ensemble
 
+        # Note on scaling
+        # -------------------
+        # In line 8 in Algorithm 1, we solve (S S^T + E E^T) X = (SW + D - g(X))
```
dafeda, line 165: Not immediately clear to me that this is what is being solved in line 8 of Algorithm 1.
Can you dumb this down a bit or do I just have to stare harder at the equations?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,66 +130,76 @@ def fit(
...
+        # for X, and then we compute S.T X. If we scale the rows (observed variables)
+        # of these equations using the standard deviations, we can obtain better
+        # conditioning numbers on the equation. This corresponds to left-multiplying
+        # with a diagonal matrix L := sqrt(diag(C_dd)). In an experiment with
```
dafeda, line 169: Nice job on testing that this improves the condition number 90% of the time 👍

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,66 +130,76 @@ def fit(
...
+        # random covariance matrices, this improved the condition number ~90%
+        # of the time.
+        # To see the equality, note that if we scale S, E and K := (SW + D - g(X))
```
dafeda, line 172: `K` here is `H` from line 7 right? Any reason not to use `H`?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,66 +130,76 @@ def fit(
...
+        # we obtain
+        #              (L S) (L S)^T + (L E) (L E)^T X_2 = L K
+        #              L (S S^T + E E^T) L X_2 = L K
+        # so the new solution X_2 := L^-1 X, expressed in terms of the original X.
+        # But when we left-multiply with S^T, we do so with a transformed S,
+        # so we obtain S_2^T X_2 = (L S)^T (L^-1 X) = S^T X, so the solution
+        # to the transformed system is equal to the solution of the original system.
```
dafeda, line 179: Najs!

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -88,66 +130,76 @@ def fit(
...
+        # In the implementation of scaling the right hand side (SW + D - g(X))
+        # we first scale D - g(X), then we scale S implicitly by solving
+        # S Sigma = L Y, instead of S Sigma = Y for S.
+
         # Scale D and E with observation error standard deviations.
         D /= observation_errors_std.reshape(num_obs, 1)
-        E /= observation_errors_std.reshape(num_obs, 1)
+        E = E / observation_errors_std.reshape(num_obs, 1)
```
dafeda, line 186: From my understanding, this creates a copy of `E`, while using the in-place operator `/=` does not. Am I wrong?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: If they are not overused, I like them! Here the method does two things: (1) housekeeping / input validation and (2) computatations -- and I like splitting it up for the reader.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: Dumbed it down a bit :)

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: Good catch. Should of course use H instead of introducing new variables!

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: You are correct! We have to copy here, since we do not draw E again and again in each iteration now. We use the same E, and then we cannot scale in each iteration.

This is a bit messy overall, I hope to fix more of it in future PRs - but for now the memory test still passes, so this approach does not use much more memory

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -85,70 +127,87 @@ def fit(
...
+        # H := (SW + D - g(X)). This is equivalent to solving the following
+        # equation for an unknown matrix M:
+        #     (S S^T + E E^T) M = (SW + D - g(X))
+        # Afterwards we compute S.T M. If we scale the rows (observed variables)
+        # of these equations using the standard deviations, we can obtain better
+        # conditioning numbers on the equation. This corresponds to left-multiplying
+        # with a diagonal matrix L := sqrt(diag(C_dd)). In an experiment with
+        # random covariance matrices, this improved the condition number ~90%
+        # of the time (results may depend on how random covariance matrices are
+        # generated --- I covariances C by first E ~ stdnorm(), then C = E.T @ E).
```
dafeda, line 175: Did you perhaps mean to write "I generated covariances C ..."?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: Not sure I follow.
Both `E = E / observation_errors_std.reshape(num_obs, 1)` and `E /= observation_errors_std.reshape(num_obs, 1)` should produce the same result but the last one will use less memory right?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
```diff
@@ -175,7 +234,7 @@ def update(self, param_ensemble: npt.NDArray[np.double]) -> npt.NDArray[np.doubl
             np.ix_(self.ensemble_mask, self.ensemble_mask)
         ]
         transition_matrix /= np.sqrt(ensemble_size - 1)
-        transition_matrix.flat[:: ensemble_size + 1] += 1.0
+        transition_matrix.flat[:: ensemble_size + 1] += 1
```
dafeda, line 237: Think I prefer 1.0 since it's more explicit.

src/iterative_ensemble_smoother/utils.py
```diff
@@ -44,12 +45,35 @@ def response_projection(
...
+    # The code below is equivalent to np.linalg.pinv(A) @ A, since:
+    # pinv(A) @ A = [V (1/s) U.T] @ [U s V.T]
+    #             = V (1/s) @ s V.T
+
+    # Compute SVD
+    U, s, VT = sp.linalg.svd(A, full_matrices=False)
+
+    # This code is adapted from
+    # https://github.com/numpy/numpy/blob/db4f43983cb938f12c311e1f5b7165e270c393b4/numpy/linalg/linalg.py#L2015
+    rcond = np.asarray(1e-15)
```
dafeda, line 70: What do we gain by doing this instead of doing `np.linalg.pinv(A) @ A`?

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: I used 1.0 since it works if the matrix is of type integer or float. If we use 1 then it will not work on integers. Not that the transition matrix is integer anyway :) can go back to 1.0 :)

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: yes, will fix!

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
tommyod: If we don't cache E, then we have to cholesky decompose the covariance matrix in every iteration, plus multiply the cholesky factor with a vector. That's not nothing, but I have not timed it. Certainly more time and memory than if we cache it (even with mutating it).

In the future it will not be mutated. E is used for 2 purposes in the algorithm: (1) to add noise to observations and (2) to create a low-rank approximation of the covariance. Actually, there is no reason why E has to have the same dimensional for these purposes. If we have 5 output variables and 2 ensemble members, then we use E of shape (5, 2) for both purposes. But using E of shape e.g. (5, 3) would be a better approximation to the covariance matrix. We can separate this clearly and document it.

In the future I do not want the code to look like it does in this PR, but I am trying to attack the problem step by step. Caching E removes some unneccessary work (the repeated cholesky), but there is still work to do

src/iterative_ensemble_smoother/utils.py
tommyod: `np.linalg.pinv(A)` essentially computes the SVD, then multiplies factors together:

```
pinv(A) @ A = V.T * 1/s @ U.t @ A
```
so we take SVD, then perform 2 matrix multipliations with large matrices.

This code computes the SVD, then performs a single matrix multiplication with smaller matrices. So it should, in theory, be faster and more memory efficient...

However, when I time it, the `pinv` code is faster:

```
timeit.timeit("U, S, VT = sp.linalg.svd(A, full_matrices=False); VT.T @ VT;", setup="import numpy as np; import scipy as sp; A = np.random.randn(10_000, 100)", number=99) / 99
Out[204]: 0.08380154609054591

timeit.timeit("np.linalg.pinv(A) @ A", setup="import numpy as np; A = np.random.randn(10_000, 100)", number=99) / 99
Out[205]: 0.06609695008132757
```

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: From my experience they usually end up outdated rather quickly, but can leave for now.

src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py
dafeda: I'm principally against adding any kind of state unless it is of considerable benefit, and I am not sure whether that is the case here.
I've not seen E used to create a low-rank approximation in the wild so I do not deem it very important.
Let's have a think and discuss via video.

src/iterative_ensemble_smoother/utils.py
dafeda: Let's avoid adding any extra complexity unless it is measurably and considerably better.

tommyod: Agree

dafeda (approved):
Great 👍

PR #124, opened 2023-08-08
Removed `ensemble_size` argument from SIES

tommyod:
The `ensemble_size` argument is used to set up the coefficient matrix. But there is no need to set it up during initialization. If we set it up on the first `fit` call instead, we don't need the argument at all, since we can deduce the ensemble size from the other arguments sent to `fit`.

This makes life easier for users. I also think it's a good principle to not include multiple input arguments if some can be deduced from others.

What do you think?

**Warning: Breaking public API change**


dafeda (approved):
This makes sense I think 👍

PR #131, opened 2023-08-15
ESMDA docs

tommyod:
- Added a notebook showcasing ESMDA on a simple example with a mutual fund simulation.
- Added reference to ESMDA paper in sphinx
- Updated docstrings for SIES (pr #110) 

PR #110 was merged into this one, since it was related to sphinx.


tommyod:
As agreed with @dafeda I removed the mutual fund example and used ESMDA on the oscillator instead.

tommyod:
@dafeda , can you look over / approve?

Issue #116, opened 2023-07-12
Why center E?


tommyod:
In this line E is centered
https://github.com/equinor/iterative_ensemble_smoother/blob/24d771149134299102f8770c397b670c1d1efbf8/src/iterative_ensemble_smoother/_iterative_ensemble_smoother.py#L116

This decreases the rank by 1.

-------
Under equation 14 it says "_E is the centered measurement-perturbation matrix whose columns are sampled from N(0, C_dd)_".
Here "centered" could be interpreted as (1) that the mean is 0 in N(0, C_dd) or that (2) the rows in E should be centered.

-------
If we think of equation (15), then the _sample covariance_ does require us to define E as mean centered and divided by (N-1).
But don't we already know the true mean to be zero (ref. our model is eqn (7))?
So the best low-rank approximation to C_dd would be 
```
# approach 1
a ~ N(0, C_dd)
C_dd_bar = a @ a.T / len(a)
```
and not
```
# approach 2
a ~ N(0, C_dd)
C_dd_bar = subtract_mean(a) @ subtract_mean(a).T / (len(a) - 1)
```
Again I appeal to the case when we draw a single sample.
Approach 1 would produce a rank 1 approximation to C_dd, while approach 2 would produce a rank 0 approximation to C_dd (C_dd_bar is identically zero).

-------

If we have a run with a single ensemble member, then E will be identically zero, and there will be no permutation of the observations and no sampling. I would expect that:

- 1000 runs with 1 ensemble member (different seeds)
- 1 run with 1000 ensemble member

should both sample the posterior, but in our implementation the first option will always give the same answer is E is identically zero.

-------

Thoughts on this @Blunde1 ? (we can discuss over chat. I just wanted to write it down here)

tommyod:
@Blunde1 [wrote](https://github.com/equinor/iterative_ensemble_smoother/pull/136#discussion_r1305294908):

![image](https://github.com/equinor/iterative_ensemble_smoother/assets/10076072/734439a5-7c7e-4751-b57d-753181bec02c)

tommyod:
After talking to @Blunde1 and @dafeda about this, none of us found any good reason to keep the centering. It was removed in #136 .

PR #136, opened 2023-08-23
Refactor SIES

tommyod:
This PR is a draft of how I would like the SIES implementation to be. It's meant to be a starting point for discussion.

**Some highlights of the code:**

- The (internal) variables names follow the reference paper exactly.
- The code is documented line by line how it follows Algorithm 1.
- The API takes all static (unchanging) parameters in the constructor `SIES(*static_args)`, and the only parameter that goes into the `newton()` call is the `Y` (`response_ensemble`). The problem with the having to maybe send in `X` (`param_emsemble`) is solved.
- This code passes the memory test that the current `SIES` implementation passes.
- This code is equally fast as the current `SIES` implementation in the case of a diagonal covariance matrix. Tested on a single iteration. Over many iterations it might be faster, since it caches the Cholesky factorization.

Since this is just meant as a starting point for discussion, **I did not**:

- Implement all inversion methods.
- Write many tests (all existing tests should be used).
- Use typing very much.
- Separate code into several files. I just used 2 files for simplicity so far.


tommyod:
Ready for a discussion. Feel free to leave some notes here, but we can also schedule a meeting :)

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,315 @@
...
+        # Line 6
+        if n < N - 1:
+            # There are fewer parameters than realizations. This means that the
+            # system of equations is overdetermined, and we must solve a least
+            # squares problem.
+
+            # An alternative approach to producing A_i would be keeping the
+            # returned value from the previous Newton iteration (call it X_i),
+            # then computing:
+            # A_i = (self.X_i - self.X_i.mean(axis=1, keepdims=True)) / np.sqrt(N - 1)
```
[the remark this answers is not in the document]
Blunde1: Perhaps we have equivalence, and we do not need to flag anything?
So it becomes a question of potential extra computation (if we could have used the final line directly) in an edge case, versus the cost of an additional parameter upon initialization (it could be optional).
image

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,315 @@
...
+        else:
+            self.C_dd_cholesky = np.sqrt(self.C_dd)
+
+        # Equation (14)
+        self.D = self.d.reshape(-1, 1) + sample_mvnormal(
+            C_dd_cholesky=self.C_dd_cholesky, rng=self.rng, size=self.X.shape[1]
+        )
+
+        self.W = np.zeros(shape=(self.X.shape[1], self.X.shape[1]))
+        self.X_i = self.X
```
Blunde1, line 50: This is not employed anywhere, I think, except in comments. Might be a remnant from some other code?
If this is intended to be used as `X_i = X + X@W_i  / np.sqrt(N-1)` ( adding subscript to `W`, since it is appropriate), then care must be taken to not accidentally overwrite `X` at some point. I think the original code had a `.copy()` to avoid this, which also is not that good. Probably best to avoid this one.

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,315 @@
...
+            W=self.W,
+            step_length=step_length,
+            S=S,
+            C_dd=self.C_dd,
+            H=H,
+            C_dd_cholesky=self.C_dd_cholesky,
+        )
+
+        # Line 9
+        return self.X + self.X @ self.W / (np.sqrt(N - 1))
```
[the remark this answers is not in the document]
Blunde1: Uncertain about the `return` statement. See general review

Blunde1, line 107: Do we need a parenthesis around `np.sqrt(N-1)`?

Blunde1 (commented):
I have yet to look at inversion functions and below.

I find the implementation readable, and easily corresponds to the 2019 reference paper. This has been an issue before, and to me seems like an improvement. 
The small detail I like the most is the calculation of the `C_dd` Cholesky factor upon initialization and reusable as an attribute. What I like the least is the name `newton`.

Some discussion around the api:
- I think we cannot get around passing `X`/`param_ensemble` in the initialization, at least optionally. To me it is remarkable that we almost do not have to do this (n>N-1 or linear model), and how it occurs (n>N-1!). It is something that would be expected in other ensemble based data assimilation algorithms (as long as some kind of regression/statistical learning problem is solved). Thus, if there are memory issues with passing `X`, then this will occur for most other algorithms.
- One could imagine running `SIES.newton()` multiple times without the return (line 9) (corresponding to "fit" in sklearn world) and then running an "SIES.update(X)" corresponding to line 9/the return statement. This would clearly separate the statistical estimation and the update, which would make it easier to inspect and report on both procedures appropriately. The downside to this, and in favor of the PR, is that it is hard to comment on the iterative statistical regression problem that is solved, so why bother? Also, the current proposal aligns more with Algorithm 1 in the paper, so point for simplicity and accuracy. The separation is perhaps more of an ideal, but if it is not possible it should perhaps be scrapped.
- Should we return updated `X` or hold the "state" `W`? Both hold the information from the iterations. Depends perhaps on what one thinks of the point above. I myself is not sure.
- It is good that `X` is not passed to `newton`. We could easily get a situation where we return `X_i + X_i @ self.W / np.sqrt(N - 1)` intended as `X_{i+1}`.
- If we aim for a pure implementation, then should we flag if the forward model is linear? I guess this would be the case if we are purists, and/or if `sp.linalg.pinv(A_i)@ A_i` is costly. If the cost is low, then I suggest we drop it, as in the PR proposal.

The basis for my thoughts is that we are aiming for a pure implementation of the algorithm, that is numerically stable and scalable.

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+# =============================================================================
+# UTILITY FUNCTIONS
+# =============================================================================
+def center(X):
+    # Center each row, in place, so sum(row) = 0 for every row
+    X -= X.mean(axis=1, keepdims=True)
```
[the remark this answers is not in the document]
tommyod: Not used anywhere. Deleting it.

Lets discuss centering E in [this issue](https://github.com/equinor/iterative_ensemble_smoother/issues/116). I agree with you - I see no reason to center E.

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+class SIES:
+    def __init__(
+        self,
+        param_ensemble: npt.NDArray[np.double],
+        observation_errors: npt.NDArray[np.double],
+        observation_values: npt.NDArray[np.double],
+        *,
+        seed: Optional[int] = None,
```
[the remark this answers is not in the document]
tommyod: Here I want to deviate a little bit from the paper.

**I think we can safely assume it's not.** If it was, then we would just use standard results for posteriors of multivariate normal distributions. Newtons steps and ensemble smoothers would be overkill for linear models anyway.

Even if it is linear, it has no cost except an unnecessary projection since $A_i^+ A_i = I$ in the linear case anyway.

src/iterative_ensemble_smoother/experimental_sies.py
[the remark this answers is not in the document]
tommyod: Agreed that it would be present upon initialization. Agree that this is a trade-off vs extra computation or extra argument.

I think we should not have the argument, since it's fair to assume that if one is running ensemble smoothers one does not have a linear model. In the toy cases where the model is actually linear, we can just do one extra projection $A_i^+ A_i = I$. It's not a bottleneck.

src/iterative_ensemble_smoother/experimental_sies.py
tommyod: Agreed. Best to avoid this one.

src/iterative_ensemble_smoother/experimental_sies.py
tommyod: nope!

tommyod:
> I have yet to look at inversion functions and below.

That's OK. They are in some ways the simple part. The API is the hard part.

>  I find the implementation readable, and easily corresponds to the 2019 reference paper. This has been an issue before, and to me seems like an improvement. 

Great!

> I think we cannot get around passing `X`/`param_ensemble` in the initialization, at least optionally.

Agreed.

> One could imagine running `SIES.newton()` multiple times without the return (line 9) (corresponding to "fit" in sklearn world) and then running an "SIES.update(X)" corresponding to line 9/the return statement. This would clearly separate the statistical estimation and the update, which would make it easier to inspect and report on both procedures appropriately.

Fantastic input @Blunde1 - this idea solves many problems, but requires some more low-level methods and some high-level methods. **I implemented an idea for this just now**. The method `propose_W` runs lines 4-8, and the method `sies_iteration` calls on it to run lines 4-9. Have a look! I also show to to implement a halving line search in `line_search` based on this idea of separation. We can rename a bit, and we can perhaps add underscores to signify "private" methods.

> If we aim for a pure implementation, then should we flag if the forward model is linear? I guess this would be the case if we are purists, and/or if `sp.linalg.pinv(A_i)@ A_i` is costly. If the cost is low, then I suggest we drop it, as in the PR proposal.

I suggest we drop it.

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,418 @@
...
+            C_dd_cholesky=self.C_dd_cholesky,
+        )
+
+
+# =============================================================================
+# UTILITY FUNCTIONS
+# =============================================================================
+
+
+def sample_mvnormal(*, C_dd_cholesky, rng, size):
```
dafeda, line 212: Why the "*" as first argument?

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,418 @@
...
+    """Draw samples from the multivariate normal N(0, C_dd).
+
+    We write this function from scratch here we can to avoid factoring the
```
dafeda, line 215: Typo: "we can" can be deleted I think.

src/iterative_ensemble_smoother/experimental_sies.py
```diff
@@ -0,0 +1,418 @@
...
+    S = S * scale_factor.reshape(-1, 1)
+    H = H * scale_factor.reshape(-1, 1)
+
+    # We define K as the solution to
+    # (S @ S.T + C_dd) @ K = H, so that
+    # K = (S @ S.T + C_dd)^-1 H
+    # K = solve(S @ S.T + C_dd, H)
+    if C_dd.ndim == 1:
+        lhs = sp.linalg.blas.dsyrk(alpha=1.0, a=S)  # S.T @ S
+        lhs.flat[:: lhs.shape[0] + 1] += 1
```
dafeda, line 347: What if we use the following instead as I think it's easier to understand?

np.fill_diagonal(lhs, lhs.diagonal() + 1)

tommyod:
Feedback:

- Keep things simple
- Get rid of lapack/blas calls unless SIGNIFICANT speedups are achieved (an order of magnitude)
- Api looks good for the most part.
- Move line search out of the method for now, but keep it in a notebook or something
- Write proper docstrings and types

tommyod:
### Review

This PR touches many lines. I think it's best if someone reviews the high level stuff (API, files, skim the tests, etc). If we agree on that we merge, and the details can be fixed in PRs in the future.


### General changes

The following changes have been done since the last discussion point:

- Types have been added.
- Several inversion methods have been added (3 of them).
- Docstrings have been written.
- All calls to function in `scipy.linalg.blas` have been removed.
- The line search method has been removed. Stored in a test for now.

### Tests

I had to review every single test and update it. In the process I structured the tests differently to accommodate `SIES` and `ESMDA`. The current structure is: 

```text
src/iterative_ensemble_smoother
├── esmda_inversion.py
├── esmda.py
├── __init__.py
├── sies_inversion.py
├── sies.py
└── utils.py
tests
├── __init__.py
├── test_esmda_inversion.py
├── test_esmda.py
├── test_sies_inversion.py
└── test_sies.py
```

Some other changes to tests:

- Many test files had global variables and functions. I refactored these into the tests.
- IMO the tests in `test_snapshots` do not test anything except the fact that outputs on specific problems did not change (regression tests). For this purpose using the CSV snapshots seems needlessly complex (can just define the output as a variable and test against it?), also the construction of non-normal variables, "complex" forward models, etc seems needlessly complex. Might as well just use a linear model and test on that for the purpose of a regression test. I have removed almost all of this. Willing to discuss it of course.
- Removed `test_that_projection_is_better_for_nonlinear_forward_model_big_N_small_n`. This test (1) depends on implementation details (do not use the class API), (2) does not test our implementation, but tests a mathematical thing from the paper.
- Simplified tests such as `test_that_sies_converges_to_es_in_gauss_linear_case` drastically. Before they used polynomial models, `p_map`, a mix between pure Python code and NumPy, etc. Now I just use a model `g = lambda x: A @ x` for testing. We just want to test on a gauss-linear model. No reason to complicate it.
- Some more general housekeeping on tests: parameterized a bit more, tested a few more properties, etc.

Memory test and theoretical tests on properties all pass.

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+
+    Parameters
+    ----------
+    parameters : npt.NDArray[np.double]
+        A 2D array of shape (num_parameters, ensemble_size). Each row corresponds
+        to a parameter in the model, and each column corresponds to an ensemble
+        member (realization). This is X in Evensen (2019).
+    covariance : npt.NDArray[np.double]
+        Either a 1D array of diagonal covariances, or a 2D covariance matrix.
+        The shape is either (num_parameters,) or (num_parameters, num_parameters).
```
dafeda, line 36: The shape here should be (num_observations,) or (num_observations, num_observations).

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+            - `subspace_exact` : Solve Eqn (42) using Eqn (50), i.e., the Woodbury
+               lemma to invert a matrix of size (ensemble_size, ensemble_size)
+            - `subspace_projected` : Solve Eqn (42) using Section 3.3, i.e.,
+               by projecting the covariance onto S. This approach utilizes the
+               truncation factor `truncation`.
+
+    truncation : float
+        How much of the total energy (singular values squared) to keep in the
+        SVD when `inversion` equals `subspace_projected`. Choosing 1.0
+        retains the most information, while 0.0 removes all information.
```
dafeda, line 60: Is it more accurate to say that choosing 1.0 retains __all__ information?

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+        """Evaluate the objective function in Equation (18), taking the mean
+        over each ensemble member.
+
+        Parameters
+        ----------
+        W : npt.NDArray[np.double]
+            Current weight matrix W, which represents the current X_i as a linear
+            combination of the prior. See equation (17) and line 9 in Algorithm 1.
+        Y : npt.NDArray[np.double]
+            Reponses when evaluating the model at X_i. In other words, Y = g(X_i).
```
dafeda, line 132: Evensen writes in equation (1) that y=g(x).
However, in equation (30) he writes Y_i = g(X_i)(I - 1/N11^T)/sqrt(N-1) which means that Y is centered.
Should we use a different symbol for non-centered and centered responses?

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+            # Evaluate the elementwise likelihood term
+            # Evaluate np.diag((g(X_i) - D).T @ C_dd^{-1} @ (g(X_i) - D))
+            #        = np.diag((Y - D).T @ C_dd^{-1} @ (Y - D))
+            # First let A := Y - D, then use the cholesky factors C_dd = L @ L.T
+            # to write (Y - D).T @ C_dd^{-1} @ (Y - D)
+            #             A.T @ (L @ L.T)^-1 @ A
+            #             (L^-1 @ A).T @ (L^-1 @ A)
+            # To compute the expression above, we solve K := L^-1 @ A,
+            # then we compute np.diag(K.T @ K) as (K**2).sum(axis=0)
+            A = Y - self.D
```
dafeda, line 180: Can we use a different symbol than `A` here?
Also, `A = Y-self.D` can be moved above the `if` since it's used in both `if` and `else`.

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+            K = sp.linalg.blas.dtrsm(alpha=1.0, a=self.C_dd_cholesky, b=A, lower=1)
```
dafeda, line 181: Can we avoid calling `blas` directly?

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+            likelihood = (K**2).sum(axis=0)
+            # assert np.allclose(likelihood, np.diag(A.T @ np.linalg.inv(self.C_dd) @ A))
```
dafeda, line 183: Think we can delete this commented out line.

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+
+        else:
+            # If C_dd is diagonal, then (L^-1 @ A) = A / L.reshape(-1, 1)
+            A = Y - self.D
+            K = A / self.C_dd_cholesky.reshape(-1, 1)
+            likelihood = (K**2).sum(axis=0)
+            # assert np.allclose(likelihood, np.diag(A.T @ np.diag(1/self.C_dd) @ A))
```
dafeda, line 190: Think we can delete this commented out line.

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+    def sies_iteration(
+        self,
+        responses: npt.NDArray[np.double],
+        step_length: float = 0.5,
+        ensemble_mask: Optional[npt.NDArray[np.bool_]] = None,
+    ) -> npt.NDArray[np.double]:
+        """Perform a single SIES iteration (Gauss-Newton step).
+
+        This method implements lines 4-9 in Algorithm 1.
+        It returns a updated X and updates internal state W.
```
dafeda, line 203: Nitpick:

"It returns an updated X and updates the internal state W."

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+        m = self.C_dd.shape[0]  # Responses (outputs)
+
+        # Line 4 in Algorithm 1
+        Y = (g_X - g_X.mean(axis=1, keepdims=True)) / np.sqrt(N - 1)
+
+        # Line 5
+        Omega = self.W.copy()
+        Omega -= Omega.mean(axis=1, keepdims=True)
+        Omega /= np.sqrt(N - 1)
+        Omega.flat[:: Omega.shape[0] + 1] += 1  # Add identity in place
```
dafeda, line 277: Your new favourite trick: `np.fill_diagonal(Omega, Omega.diagonal() + 1)` :)

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+
+        # Line 6
+        if n < N - 1:
+            # There are fewer parameters than realizations. This means that the
+            # system of equations is overdetermined, and we must solve a least
```
dafeda, line 282: Fewer parameters than unknowns is underdetermined right?

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+        -------
+        W_i : npt.NDArray[np.double]
+            A proposal for a new W in the algorithm.
+
+        """
+        Y = responses
+        assert Y.ndim == 2
+        assert Y.shape[0] == self.C_dd.shape[0]
+        g_X = Y.copy()
+        if ensemble_mask is not None:
```
dafeda, line 346: `ensemble_mask` can not be `None` in this function I think

src/iterative_ensemble_smoother/sies.py
```diff
@@ -0,0 +1,416 @@
...
+
+            # An alternative approach to producing A_i would be keeping the
+            # returned value from the previous Newton iteration (call it X_i),
+            # then computing:
+            # A_i = (self.X_i - self.X_i.mean(axis=1, keepdims=True)) / np.sqrt(N - 1)
+            A_i = self.A @ Omega
+            # S = sp.linalg.solve(
+            #     Omega.T, np.linalg.multi_dot([Y, sp.linalg.pinv(A_i), A_i]).T
+            # ).T
+            ST, *_ = sp.linalg.lstsq(
```
dafeda, line 379: Why not use `sp.linalg.solve` here?

src/iterative_ensemble_smoother/sies_inversion.py
```diff
@@ -0,0 +1,379 @@
...
+    W: npt.NDArray[np.double],
+    step_length: float,
+    S: npt.NDArray[np.double],
+    C_dd: npt.NDArray[np.double],
+    H: npt.NDArray[np.double],
+    C_dd_cholesky: Optional[npt.NDArray[np.double]],
+    truncation: float,
+) -> None:
+    """Verify shapes of arguments."""
+    if C_dd_cholesky is not None:
```
dafeda, line 121: Should not be `None` based on type.

src/iterative_ensemble_smoother/sies_inversion.py
```diff
@@ -0,0 +1,379 @@
...
+    # Scale rows
+    S = S * scale_factor.reshape(-1, 1)
+    H = H * scale_factor.reshape(-1, 1)
+
+    # We define K as the solution to
+    # (S @ S.T + C_dd) @ K = H, so that
+    # K = (S @ S.T + C_dd)^-1 H
+    # K = solve(S @ S.T + C_dd, H)
+    if C_dd.ndim == 1:
+        lhs = S @ S.T  # sp.linalg.blas.dsyrk(alpha=1.0, a=S)
```
dafeda, line 211: Can we get rid of the comment refering to dsyrk?

src/iterative_ensemble_smoother/sies.py
tommyod: write extra sentence to clarify this. but leave it as Y unless we figure out something better.

src/iterative_ensemble_smoother/sies.py
tommyod: i think there might have been a bug earlier where we said that `W` had to be `W[mask, mask]`. but if we start out with e.g. 10, and we end up with 5, that does not mean that the remaining 5 have to be linear combinations of the ORIGINAL 5. I think `W[:, mask]` is correct.

tommyod: i think there might have been a bug earlier where we said that `W` had to be `W[mask, mask]`. but if we start out with e.g. 10, and we end up with 5, that does not mean that the remaining 5 have to be linear combinations of the ORIGINAL 5. I think `W[:, mask]` is correct.

dafeda (approved):
Very nice!

PR #139, opened 2023-09-01
Updated Oscillator.py and Polynomial.py

### LINKED ISSUE #137: Examples that show sampling more clearly
The current examples in the docs have so low measurement noise $C_{dd}$ that the posterior samples almost appear to converge to single numbers.

I would be interesting if the examples:

1. Showed a more wide posterior, where the posterior is not so dominated by the low measurement noise
2. Showed the multivariate posterior (e.g. scatterplot in parameter space) instead of just marginal histograms, to highlight that the posterior can have interesting shapes

tommyod:
Closes #137

### Oscillator.py

- Added some $\LaTeX$ explaining the anharmonic oscillator equations
- Found an off-by-one error in the implementation of the anharmonic oscillator. I re-wrote the equation away from the transition matrix, to a pure implementation of the difference equation. I think it is easier to understand, and less code.
- Vectorized a loop or two
- I added more observational noise, to show how the posterior is a tradeoff between the prior and likelihood. If the observational noise is too low, the algorithm looks more like an optimization routine than (approximate) posterior sampling.

**Question** I refactored sampling. Instead of sampling uniform and transforming to normal, I just sample normal variables and use those. **This changes the example somewhat.** I did it because I am not sure if sampling from uniform is valid. If my prior is `x ~ U(0, 1)`, then I am saying anything between 0 and 1 is equally likely. If I transform those samples to a normal distribution, then my prior says that 0.5 is roughly the most likely. But that's not the same assumption! Perhaps residing statistician @Blunde1 tell me if my intuition is right or wrong.

Added some extra figures, such as this one:
image

### Polynomial.py

- I added more observational noise, to show how the posterior is a tradeoff between the prior and likelihood. If the observational noise is too low, the algorithm looks more like an optimization routine than (approximate) posterior sampling.

Added some extra figures:

image

image

### DIFF
--- a/docs/source/Oscillator.py
+++ b/docs/source/Oscillator.py
@@ -87,7 +87,7 @@ def plot_result(A, response_x_axis):
 # ## Setup
 
 # %%
-def _generate_observations(K):
+def generate_observations(K):
     """Run the model with true parameters, then generate observations."""
     # Evaluate using true parameter values on a fine grid with K sample points
     x = simulate_anharmonic(omega=3.5e-2, lmbda=3e-4, K=K)
@@ -125,7 +125,7 @@ def forward_model(A, response_x_axis):
 
 
 response_x_axis = range(2500)
-observation_values, observation_errors, observation_x_axis = _generate_observations(
+observation_values, observation_errors, observation_x_axis = generate_observations(
     len(response_x_axis)
 )
 
@@ -154,8 +154,8 @@ plt.show()
 
 # %%
 realizations = 100
-priors = [(2.5e-2, 4.5e-2), (2.0e-4, 4.0e-4)]
-A = np.vstack([rng.uniform(low, high, size=realizations) for (low, high) in priors])
+priors = [(0.035, 0.01), (0.0003, 0.0001)]
+A = np.vstack([rng.normal(low, high, size=realizations) for (low, high) in priors])
 
 plot_result(A, response_x_axis)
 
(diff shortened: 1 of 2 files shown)


Blunde1:
> **Question** I refactored sampling. Instead of sampling uniform and transforming to normal, I just sample normal variables and use those. **This changes the example somewhat.** I did it because I am not sure if sampling from uniform is valid. If my prior is `x ~ U(0, 1)`, then I am saying anything between 0 and 1 is equally likely. If I transform those samples to a normal distribution, then my prior says that 0.5 is roughly the most likely. But that's not the same assumption! Perhaps residing statistician @Blunde1 tell me if my intuition is right or wrong.

Sampling from a uniform / a uniform prior is valid. However, doing the update in this space is not valid. The update is valid for a sample from a normal prior, thus the transformation to this space (and transformation back when presented to users) before the update. This comes at the cost of more non-linearity between `x` and `y` learned by the (implicit) linear regression, but is likely better than updating in a non-suitable space, even if the map indeed was linear.

tommyod:
Thanks for the patience @Blunde1 . I understand it now, and I have added an extensive comment to document what is going on in the notebook :)

tommyod:
Ready for a review. I think this should be relatively uncontroversial, but let me know if anyone disagrees with anything.

dafeda (approved):
Looks great 👍

PR #140, opened 2023-09-06
Remove requirements.txt in readthedocs

tommyod:
- Readthedocs now uses similar approach to tox / GH action that builds docs.
- In the process we remove an unneccesary file.


dafeda (approved):
This may cause our docs to fail more often so not sure if it's a great idea, but let's try it and see how goes.

PR #141, opened 2023-09-14
Refactoring code + update the documentation

antoinecollet5:
This is a surface refactoring with the following improvements:

* Documentation
   * Use pydata template instead of rst
   * Include the readme in the doc for a nice rendering and avoid duplication
   * Make links between the bibliography and the docstrings using the :cite: directive.
   * Better API section

* Precommit
  * Add ruff for linting
  * Add isort for imports autofix
  * Add codespell to avoid typos

*  SIES interface: you might disagree, but in python, using `hasattr` is usually a bad practice and means problem with the design (see #124). Typically, I found it more logical to create the observation perturbations (`E_`) in the `__init__` method rather than somewhere in the `fit` method (which by the way had to much responsibility). In addition, it allows to create nice properties (`D`, `num_obs`, ...) which makes the code easier to read and to test. There is technically no reason to change `observation_values`,  nor `observation_errors` while calling `fit` multiple times, so passing it once at the instance creation seems a better idea.
Using `__slots__` you make sure that all class attributes are well initialized at the instance creation, and you also save some memory. Note that we should probably update the class `ES` which I did not.

* Re-establish the `InversionType` class (removed in #101). I named it `SiesInversionType` because I guess the same can be done with ESMDA. The trick here is to use a str enum type. I believe everybody is happy since you can still use string such as `naive` which does no break any code, or prefer SiesInversionType.NAIVE if you want to ensure strong typing in your code (and auto-completion support). I added some testing.

* Makefile: introduction of a makefile at the root of the project. You can now run a simpler commands to build, test, etc. 

```
> make

clean                remove all build, test, coverage and Python artifacts
clean-build          remove build artifacts
clean-pyc            remove Python file artifacts
clean-test           remove test and coverage artifacts
lint                 check style with flake8
test                 run tests quickly with the default Python
test-all             run tests on every Python version with tox
coverage             check code coverage quickly with the default Python
docs                 generate Sphinx HTML documentation, including API docs
servedocs            compile the docs watching for changes
release              package and upload a release
dist                 builds source and wheel package
install              install the package to the active Python's site-packages
```


tommyod:
There are some great ideas and work here! The enthusiasm is very much appreciated! 

## Splitting up PRs

It's better to split this work up into several PRs. Sometimes it's unavoidable to create large PRs, but it's nice if each PR solves one issue at a time - especially when the issues are independent of each other. The way I see, it, this PR addresses a few different issues that are completely independent of each other:

- Documentation improvements
- Code quality improvements (isort, linting, etc)
- Code changes (inversion type)
- etc

I propose that you split this up into several PRs, so we can review and merge each one in turn. Each PR should solve one problem or improve on one thing. 

## Create issues first

It's also better to raise issues first, so we can have a quick discussion, to avoid doing duplicate work, for instance:

- We very recently decided to get rid of the `InversionType`, since using strings instead of enums is more typical for scientific code (numpy, scipy, sklearn, etc). Reverting the decision and going back to an enum could be a possibility, but then we should discuss and decide before doing the work.
- Regarding the SIES interface - I am rewriting the entire thing in #136. This PR conflicts with that work. Again it would be better to discuss it first by raising issues or joining the discussion in that PR.

I hope this is not too discouraging. Many good ideas here.

antoinecollet5:
@tommyod thanks for the nice feedback. I created some issues and I will do my best to split the changes.

I will start by non conflicting parts.

Best

Antoine

Issue #145, opened 2023-09-20
Add ruff, isort, mypy and codespell to pre-commit hooks


antoinecollet5:
I think it could be nice to add the following pre-commit hooks:

- ruff for linting: An extremely fast Python linter, written in Rust (does the same as flake8) -> https://pypi.org/project/ruff/
- isort for imports autofix -> https://github.com/PyCQA/isort
- codespell to spot typos -> https://github.com/codespell-project/codespell

Open to discussion:
- mypy -> it is currently in the tox.ini but it could be nice to add it to the pre-commit hooks as well (double safety). What do you think ?

PR #149, opened 2023-09-20
IMP: add ruff and codespell  to pre-commit hooks

antoinecollet5:
See issue #145 
IMP: add ruff, codespell and isort to pre-commit hooks + do required corrections to comply with the tools

Note: I haven't add mypy since it is already covered by tox. But it could be an option in a second PR.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,9 @@
-[flake8]
-ignore = E302, W503, E501, E741, E203, F405
-max-line-length = 88
+[codespell]
+skip = *.pyc,*.gif,*.png,*.jpg,*.ply, ./bibliography.bib,*.ipynb
+ignore-words-list = lod,byteorder,flem,parm,doubleclick,revered,inout,fro,nd,sies,hist,ans
+quiet-level = 3
+
+[pylint.LOGGING]
+# Format style used to check logging format string. `old` means using %
+# formatting, `new` is for `{}` formatting,and `fstr` is for f-strings.
+logging-format-style=fstr
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,20 @@ repos:
                 ^docs
             )
         args: [--strict, --ignore-missing-imports, --no-warn-unused-ignores]
+
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.1.0
+    hooks:
+    -   id: codespell
+        files: ^.*\.(py|c|h|md|rst|yml)$
+        args: [
+        "docs tests",
+        "*.py *.rst *.md",
+        ]
+
+-   repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.0.289
+    hooks:
+    -   id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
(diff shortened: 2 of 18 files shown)


setup.cfg
```diff
@@ -1,3 +1,16 @@
 [flake8]
 ignore = E302, W503, E501, E741, E203, F405
 max-line-length = 88
+
+[codespell]
+skip = *.pyc,*.gif,*.png,*.jpg,*.ply, ./bibliography.bib,*.ipynb
+ignore-words-list = lod,byteorder,flem,parm,doubleclick,revered,inout,fro,nd,sies,hist,ans
+quiet-level = 3
+
+[tool.isort]
```
dafeda, line 10: Do we need `isort` when using `ruff`? Based on my understanding, `ruff` replaces `isort`.

antoinecollet5: @dafeda indeed I was not aware that ruff now includes isort. I made the changes and squashed my commits.

tommyod (approved):
Two small concerns:

- Maintaining codespell's `ignore-words-list` might become a pain. But I'm very willing to test out codespell. If it ends up being a pain in the future, we could always remove codespell from the pre commit and just run it manually from time to time.
- The cost of having two formatters instead of one. Ruff seems like it does relatively little extra compared to black, but again - let's try it out. It has 18k stars compared to blacks 34k, so it's very popular.

Overall I'm positive to merge and try this out :)

dafeda:
> Two small concerns:
> 
> * Maintaining codespell's `ignore-words-list` might become a pain. But I'm very willing to test out codespell. If it ends up being a pain in the future, we could always remove codespell from the pre commit and just run it manually from time to time.
> * The cost of having two formatters instead of one. Ruff seems like it does relatively little extra compared to black, but again - let's try it out. It has 18k stars compared to blacks 34k, so it's very popular.
> 
> Overall I'm positive to merge and try this out :)

`ruff` is not meant to replace `black` but instead `flake8`, `isort` and a bunch of others.
Here are the "Rules" `ruff` supports: https://docs.astral.sh/ruff/rules/

dafeda:
@antoinecollet5 I suggest we get rid of flake8 now that we have ruff. Do you agree?

antoinecollet5:
@dafeda I agree that flake8 is no longer needed. Do you mean removing the config from setup.cfg ?

And indeed, ruff is not overlapping with black yet. I think there are some experimental versions including black. So somewhere in the future we might remove black as well.

tommyod:
> @dafeda I agree that flake8 is no longer needed. Do you mean removing the config from setup.cfg ?

I think that's what he means. Searching [flake8](https://github.com/search?q=repo%3Aequinor%2Fiterative_ensemble_smoother%20flake8&type=code) in the repo gives no other results.

dafeda:
> > @dafeda I agree that flake8 is no longer needed. Do you mean removing the config from setup.cfg ?
> 
> I think that's what he means. Searching [flake8](https://github.com/search?q=repo%3Aequinor%2Fiterative_ensemble_smoother%20flake8&type=code) in the repo gives no other results.

Yes 👍

antoinecollet5:
@dafeda @tommyod  Done for flake8

dafeda (approved):

dafeda:
Thanks @antoinecollet5 , good work 👍

Issue #147, opened 2023-09-20
Add makefile


antoinecollet5:
I propose to add a makefile at the root of the project with the following make commands:

```
> make

clean                remove all build, test, coverage and Python artifacts
clean-build          remove build artifacts
clean-pyc            remove Python file artifacts
clean-test           remove test and coverage artifacts
lint                 check style with pre-commit hooks
test                 run tests quickly with the default Python
test-all             run tests on every Python version with tox
coverage             check code coverage quickly with the default Python
docs                 generate Sphinx HTML documentation, including API docs
servedocs            compile the docs watching for changes
release              package and upload a release
dist                 builds source and wheel package
install              install the package to the active Python's site-packages
```

tommyod:
I've been thinking about this a bit and decided that I am (weakly) against it. 

I'll try to explain why. This is the kind of thing that seems like a good idea at first, but in my opinion it's better to just learn the underlying commands. I'm in favor of a makefile if the build process is complicated, but if we're just running some tests and linters then a makefile creates another layer of abstraction that doesn't generalize to other projects and potentially prevents devs from learning the underlying commands.

As a single developer, having a file like this around can be great. On the project level it has downsides: the things mentioned above, another file to maintain, possibility that the makefile does not sync with the GH actions, etc etc. All that being said it has upsides too. It's not a bad idea at all. I'm just personally against it.

antoinecollet5:
Sure I understand your point of view. If that's okay with you I will add it to .gitignore then.
Antoine

PR #150, opened 2023-09-20
ENH: Add Makefile to .gitignore

antoinecollet5:
Associated issue: #147 
- Add makefile at the root of the project to automate usual actions required in dev phases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,11 @@ __pycache__/
 # C extensions
 *.so
 
-# Exclude the _verison.py from the git control
+# Exclude the _version.py from the git control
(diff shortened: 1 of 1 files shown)


antoinecollet5:
I don't really get why mypy keep checking _version.py ...

dafeda:
Do we agree that this can be closed based on the discussion around `Makefiles`?

antoinecollet5:
Since you do not want Makefile at the root of the project, I added it to .gitignore if you are OK with that.

dafeda:
Perhaps you can instead add it to your local .gitignore file? I don't see a reason for ignoring it in the repo to be honest.

PR #151, opened 2023-09-21
FIX: move mypy typing check to pre-commit hooks

antoinecollet5:
- FIX: move mypy typing check to pre-commit hooks to avoid having the version.py file
- TEST: merge lint.yml and typing.yml to a unique file precommit.yml -> triggered on both push and pull_requests

See issue #145 

Note: this you fix the failure in pull requests #136 , #150, #149 @tommyod


tommyod (approved):
Looks good to me!

This moves mypy into the pre-commit. Since we're already using pre-commits it does not introduce anything new. We also already test that pre-commits pass in the GH actions. This to me is a slightly more clean way of doing things. Will help catch type errors in the pre-commit stage, and merges the linting and typing GH actions into one.

Any opinion @dafeda ? I'm (slightly) in favor of this.

dafeda:
> Looks good to me!
> 
> This moves mypy into the pre-commit. Since we're already using pre-commits it does not introduce anything new. We also already test that pre-commits pass in the GH actions. This to me is a slightly more clean way of doing things. Will help catch type errors in the pre-commit stage, and merges the linting and typing GH actions into one.
> 
> Any opinion @dafeda ? I'm (slightly) in favor of this.

I'm perhaps fine with it in this small project since `mypy` runs fast, but I'm in general against slow pre-commits.

dafeda:
Would you mind squashing your commits?

tommyod:
I'll just squash-merge for him :)

PR #152, opened 2023-09-27
Float32 inputs lead to float32 outputs

### LINKED ISSUE #143: Test that using float32 inputs reduces memory usage
@dafeda introduced an obvious way to save memory: use float32 instead of float64

But this might be in vain if our implementation creates float64 matrices under the hood.
We could write a few tests to make sure that:

- If float64 is passed as inputs, the algorithms use float64 all the way through
- If float32 is passed as inputs, the algorithms use float32 all the way through

tommyod:
This closes #143

- For ESMDA, no change was needed
- For SIES, we had to make sure `E` and `W` had the correct dtype
- Regression tests are written for both ESMDA and SIES
- Neither Float16 nor Float128 is supported by linalg, so neither currently works.


dafeda (approved):
Nice 👍

PR #155, opened 2023-10-04
DOCS: improved documentation

antoinecollet5:
Several changes I propose for the documentation:

Documentation:

-    Use the modern pydata template rather than rtd (used by mainstream packages such as numpy and scipy)
-    Include the readme in the doc for a nice rendering and avoid duplication
-    Make links between the bibliography and the docstrings using the :cite: directive.
-    Improved docstrings for SIES and ES (differentiate class attributes for method parameters, detail inversion types) .
-    Better API section
-    Add makefiles for documentation build (both unix and windows) because tox does not work offline.
-    Add badges in readme for better visibility


tommyod:
Will look over this more closely soon. :) We just merged a big overhaul of SIES, leading to some conflicts unfortunately. You can probably just merge main and checkout ``--theirs` to solve the conflicts without too much trouble.

tommyod:
Hi. Fantastic work as always! Mostly looks good to me. A few comments:

- You'll have to merge in our recent changes unfortunately.
- We want to avoid makefiles. Can you elaborate on why they are needed if they are? If they are not needed, then we should remove them from the PR.

I'll also try to run it locally before we merge, just to see what the docs look like. But I'll wait until you have merged main :) Good stuff. Thanks for the effort. Greatly appreciate it.

antoinecollet5:
@tommyod thanks for the feedbacks :) I found a bit of time to solve the conflicts and adapt to the new interface. Note that I removed the example in the readme since it is now outdated.

Regarding the  Makefiles, I just removed them.

Wait for your feedback about the html rendering. Note that I did not take care of the pdf export.

docs/source/_templates/autosummary/module.rst
```diff
@@ -0,0 +1,61 @@
+{{ fullname | escape | underline}}
```
tommyod, line 1: Where is this file from? It is auto-generated from some system? Could you add a reference/source as a comment on the top?

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
+#!/usr/bin/env python
+#
+# pyesmda documentation build configuration file, created by
```
tommyod, line 3: Remove reference to `pyesmda` please :) Perhaps we should run this setup again, it if it was run in 2017 (many years ago)?

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
...
+import datetime
+
+# FIX: https://github.com/mgaitan/sphinxcontrib-mermaid/issues/72
+import errno
+import os
+import sys
 from subprocess import check_output
 
+import sphinx.util.osutil
+from sphinx.ext.napoleon.docstring import GoogleDocstring
```
tommyod, line 29: We have decided on NumPy docstrings, so I think all references to Google docstrings can go away.

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
...
+# convert the python file to a notebook
 check_output(["jupytext", "Polynomial.py", "-o", "Polynomial.ipynb"])
 
+# Do the same for this file
 check_output(["jupytext", "Oscillator.py", "-o", "Oscillator.ipynb"])
 
-
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = [
```
tommyod, line 64: Anything here that we do not need? This list grew quite a bit.

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
...
+    "sphinx.ext.autosectionlabel",  # Allow reference sections using its title
+    # allows BibTeX citations to be inserted into documentation generated by Sphinx
     "sphinxcontrib.bibtex",
+    "sphinx.ext.viewcode",  # Add links to highlighted source code
+    "sphinx.ext.intersphinx",  # Link to other projects’ documentation
+    "sphinx.ext.autosectionlabel",  # Allow reference sections using its title
+    "jupyter_sphinx",
     "numpydoc",
+    "m2r2",  # compatibility with markdown
+    "myst_nb",  # to include jupyter nteboo
```
tommyod, line 81: I note that https://github.com/spatialaudio/nbsphinx has been replaced with https://github.com/executablebooks/MyST-NB. I don't mind, if we have a reason for it.

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
...
 ]
-bibtex_bibfiles = ["refs.bib"]
+
+suppress_warnings = [
+    "nbsphinx",
```
tommyod, line 85: No need to suppress warnings if it's not used any more?

docs/source/conf.py
```diff
@@ -1,26 +1,258 @@
...
+# Solution:
+# https://michaelgoerz.net/notes/extending-sphinx-napoleon-docstring-sections.html
+# -- Extensions to the  Napoleon GoogleDocstring class ---------------------
+
+
+def parse_keys_section(self, section):
+    return self._format_fields("Keys", self._consume_fields())
+
+
+GoogleDocstring._parse_keys_section = parse_keys_section
```
tommyod, line 231: Remove google stuff, since we use NumPy docstrings.

docs/source/index.rst
```diff
@@ -1,6 +1,4 @@
-===========================
-Iterative Ensemble Smoother
-===========================
+.. mdinclude:: ../../README.md
```
tommyod, line 1: Very cool! I love the re-use of the README.

src/iterative_ensemble_smoother/sies.py
```diff
@@ -17,50 +17,61 @@
...
-    Efficient Implementation of an Iterative Ensemble Smoother for
-    Data Assimilation and Reservoir History Matching
-    written by Evensen et al (2019),
-    URL: https://www.frontiersin.org/articles/10.3389/fams.2019.00047/full
+    Note
+    ----
+    The class attributes follow the paper nomenclature.
 
-    Parameters
+    Attributes
```
tommyod, line 31: I think we should avoid "exposing" attributes in the docs. They are meant for internal use by the class and could in principle change names. Users should interact via the API, not via attributes (unlike e.g. sklearn models, where users are also encouraged to get e.g. `coefs_` from a linear model).

At least that is my thinking. I think exposing attributes leads to more noise, more stuff to maintain, more backwards compatibility issues, etc.

tommyod (commented):
Overall very excited about this. New theme looks great. Many improvements here. I do have some comments however :)

New theme looks good:

image


But LaTeX does not render:

image

Adding `"sphinx.ext.mathjax"` to the `extensions` in `conf.py` did not help me locally.

- Could you take a look at LaTeX? Does it work for you?
- I added some comments along the way, please take a look.
- In the future, even smaller PRs is good. For instance, I am not so sure about documenting attributes, but willing to discuss it. However, it's not really related to the issue of updating sphinx etc - so it would've been even better to have a PR for documentation changes (unless they are strictly required to make sphinx run).

docs/source/conf.py
antoinecollet5: Good catch :)

docs/source/_templates/autosummary/module.rst
antoinecollet5: The three files in the _template/autosummary/ folder actually define how classes, modules etc. render with autosummary. It is not automatically generated. I can add a readme at the root of the folder if you judge is necessary.

docs/source/conf.py
antoinecollet5: Fixed. I cleaned up the conf.py

docs/source/conf.py
antoinecollet5: These stuff were simply to rename the "Variable" section to "Attributes" but since you think exposing attributes might not be a good idea, this is not required.

docs/source/conf.py
antoinecollet5: There were duplicates. I cleaned it.

docs/source/conf.py
antoinecollet5: Well... I did not manage to make nbsphinx work. Don't really understand why. I usually rely on this lib. Myst works fine too. I added some config so that the maths render correctly.

docs/source/conf.py
antoinecollet5: Removed.

docs/source/index.rst
antoinecollet5: Cool :)

src/iterative_ensemble_smoother/sies.py
antoinecollet5: Ok I removed it. And you are right, this is anyway  to discuss in another PR anyway.

antoinecollet5:
@tommyod I had a quick look at latex and indeed it does not render. I propose you to see that in another issue/PR ? And to discuss whether the latexpdf export is really needed ?

docs/source/_templates/autosummary/module.rst
tommyod: If you copied it from some other repo or source, then please add a line saying so. If you wrote it yourself, then no reason to add anything.

tommyod:
> And to discuss whether the latexpdf export is really needed ?

The latexpdf is not needed. But math should render in the HTML. For me it does not render correctly in HTML (see screenshot above, same now as before). Does it render in the HTML docs for you?

I see that [m2r2 expects inline math as](https://github.com/CrossNox/m2r2/issues/20) `` `$e=mc^2$` ``, with the backticks. We use `$e=mc^2$` in our markdown. For blocks of math we use `$$e=mc^2$$` in the notebooks. I am not sure what m2r2 excepts here, or if it supports it at all.

> I propose you to see that in another issue/PR ?

I know it's tedious, but I'd rather get HTML math to work in this PR. I don't want to merge PRs that improve some things, but regress on others, and think that we'll solve it in a future PR. This kind of thinking has a tendency to lead to messy code, especially when it's not obvious what the fix is, who will dedicate time to fix it, etc.

antoinecollet5:
@tommyod normally, math should now render correctly in the html export. Have you tried since yesterday's changes ?

tommyod:
Maybe I did something wrong. Let me try again and report back ... :)

docs/source/_templates/autosummary/module.rst
antoinecollet5: These three files have been modified from the default templates provided by autosummary. See https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/autosummary.html, section "Customizing templates".
The original files can be found here: https://github.com/sphinx-doc/sphinx/tree/master/sphinx/ext/autosummary/templates/autosummary

I will add the reference in each file.

tommyod:
Fresh try:

```
git remote add antoine https://github.com/antoinecollet5/iterative_ensemble_smoother.git
git fetch --all
git switch feature/docs
pip install -e '.[dev,doc]'
python -m ipykernel install --user
sphinx-build -c docs/source/ -b html docs/source/ docs/build/html -W
firefox docs/build/html/index.html
```

And it works and looks great. :+1:

tommyod (approved):
LGTM. Thanks again for the work @antoinecollet5 . Very much appreciated as always!

PR #157, opened 2023-10-06
Rename C_D to covariance in ESMDA API

tommyod:
- Aligning the signatures of ESMDA and SIES. Both now take `covariance` as an argument.

This PR is just a search and replace.

### DIFF
--- a/src/iterative_ensemble_smoother/esmda.py
+++ b/src/iterative_ensemble_smoother/esmda.py
@@ -42,12 +42,18 @@ class ESMDA:
 
     Parameters
     ----------
-    covariance : np.ndarray
-        Covariance matrix of outputs of shape (num_outputs, num_outputs).
-        If a 1D array is passed, it represents a diagonal covariance matrix.
+    covariance : npt.NDArray[np.double]
+        Either a 1D array of diagonal covariances, or a 2D covariance matrix.
+        The shape is either (num_observations,) or (num_observations, num_observations).
+        This is C_D in Emerick (2013), and represents observation or measurement
+        errors. We observe d from the real world, y from the model g(x), and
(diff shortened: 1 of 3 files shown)


src/iterative_ensemble_smoother/esmda.py
```diff
@@ -42,7 +42,7 @@ class ESMDA:
 
     Parameters
     ----------
-    C_D : np.ndarray
+    covariance : np.ndarray
         Covariance matrix of outputs of shape (num_outputs, num_outputs).
         If a 1D array is passed, it represents a diagonal covariance matrix.
     observations : np.ndarray
```
Blunde1, line 48: I am not opposed to this change if we, in this docstring, explain its relation to the measurement error.
The SIES is brilliantly documented as
```
covariance : npt.NDArray[np.double]
        Either a 1D array of diagonal covariances, or a 2D covariance matrix.
        The shape is either (num_observations,) or (num_observations, num_observations).
        This is C_dd in Evensen (2019), and represents observation or measurement
        errors. We observe d from the real world, y from the model g(x), and
        assume that d = y + e, where the error e is multivariate normal with
        covariance given by covariance.
```

Blunde1 (commented):
My default was to be against this because there are multiple covariances going around in Kalman-type updates. I therefore wanted to build association into the naming. However, the covariance of the measurement error is the only one that is passed as input, as pointed out, so it is perhaps okay. It makes perhaps also sense to step away from the notation in the 2019 paper of SIES, if other algorithms, like ES-MDA (that have different notation in their respective papers), are implemented. A descriptive naming like `covariance` over `C_D` then makes sense. My preference is perhaps still `cov_d`, but given a good docstring I will be okay.

Blunde1 (approved):

PR #158, opened 2023-10-09
Add back row-scaling

### LINKED ISSUE #156: Add back row scaling
PR #136 removed row scaling. But ERT still uses it, so we should add it back. In the process, we should document it better.

The old code is found at:

https://github.com/equinor/iterative_ensemble_smoother/blob/627dc844bed3dfa00c4fe5c60894baf6904ff588/src/iterative_ensemble_smoother/experimental.py

tommyod:
Closes #156

### References

- Old Python code: https://github.com/equinor/iterative_ensemble_smoother/blob/627dc844bed3dfa00c4fe5c60894baf6904ff588/src/iterative_ensemble_smoother/experimental.py
- CPP code: https://github.com/equinor/ert/blob/963f9bc08ebc87374b7ed3403c8ba78c20909ae9/src/clib/lib/enkf/row_scaling.cpp#L51

### Comments

- I've rewritten the arguments to correspond to the rest of the public API.
- With the new ESMDA method `compute_transition_matrix`, which produces a matrix $K$, row scaling simply becomes

$$X_{\text{posterior}} = X_{\text{prior}} + \alpha X_{\text{prior}} K$$
- I have written the function `ensemble_smoother_update_step_row_scaling` to do a single ESMDA update (equivalent to ES) only, since that's the way it was written before. It's only a few lines of code, so from this implementation we can easily see how to extended it to multiple updates.
- I am unsure of the role of `RowScaling`, since earlier it was more or less an empty class. I _think_ the purpose of it is to hold different values for $\alpha$. I've implemented this.

Since the ESMDA method `compute_transition_matrix` makes row scaling almost trivial, we could discuss whether it belongs in `iterative_ensemble_smoother` or in `ERT`, but for now I am happy to put it back in.

**Reviewers** Please see if this implementation corresponds to your understanding of row scaling.

--------------

Unrelated change, needed to silence a warning in docs build:
```"navigation_with_keys": False```


tommyod:
The C++ row scaling code uses

$$X_{\text{posterior}} = X_{\text{prior}} F$$

instead of 

$$X_{\text{posterior}} = X_{\text{prior}} + X_{\text{prior}} K$$

which was what I assumed at first. Bad assumption. Fixed now + more documentation.

dafeda (approved):

PR #161, opened 2023-10-19
Removed 'ensemble_mask' from ESMDA

tommyod:
This pull request removes the `ensemble_mask` argument from ESMDA.

- The argument is needed is SIES, since it keeps internal state (the matrix $W$), and this must be updated correctly when we assimilate data.
- However, in ESMDA there is no such internal state. So the user can just apply the boolean mask `ensemble_mask` before sending $X$ and $Y$ to the method.

**Before** - the method is responsible for masking

```python
smoother.assimilate(X, Y, ensemble_mask=ensemble_mask)
```

**After** - the caller is responsible for masking
```python
X, Y = X[:, ensemble_mask], Y[:, ensemble_mask]
smoother.assimilate(X, Y)
```

This removes complexity in the method, letting it focus on what it's supposed to do. It also removes an argument. I see no reason for the filtering to happen in the method call, and not before.

### DIFF
--- a/src/iterative_ensemble_smoother/esmda.py
+++ b/src/iterative_ensemble_smoother/esmda.py
@@ -22,7 +22,7 @@ https://helper.ipam.ucla.edu/publications/oilws3/oilws3_14147.pdf
 
 """
 import numbers
-from typing import Optional, Union
+from typing import Union
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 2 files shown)


Blunde1 (approved):
Agree that functionality of `ensemble_mask`, if possible, should lie with the user.
LGTM

Issue #160, opened 2023-10-18
Reduce memory requirements


tommyod:
Here is an idea to reduce memory requirements.
It works, but it makes the code a bit more complicated on the low-level.
The high level API stays the same.
There is a trade-off between (1) keeping it "dumb and simple" and (2) a reduction in memory usage.
If we need it, I can implement it. If we don't need it, it seems like an overcomplication.

------------------------------------------

## Reduce memory requirements

Recall that the main computation is

$X_{i+1} = X_i + \underbrace{\text{center}(X_i) \text{center}(Y)^T /(N-1)}\_{C_{MD}}   (C_{DD} + \alpha C_D)^{-1}  (D - Y)$

If we're allowed to overwrite $X$ to save memory, then we can use:

$X += \text{center}(X) \text{center}(Y)^T /(N-1)   (C_{DD} + \alpha C_D)^{-1}  (D - Y)$

Since $X$ is the largest matrix in the typical case, we can save memory if we don't explicitly form $\text{center}(X)$ as a new matrix in memory. Instead we subtract the means, compute the right hand side, add it to X, then add back the means:

```
X_parameter_mean = np.mean(X, axis=1, keepdims=True)
X -= X_parameter_mean
X += inversion(X, Y, D, alpha, ....) # compute C_MD @ inv(C_DD + alpha * C_D) @ (D - Y)
X += X_parameter_mean
```

This avoid forming a copy of $X$ to center it. 

## Rough estimation of effects on memory usage

ESMDA memory usage will be down to 47.3 MB, compared to SIES at 70 MB. This is a reduction by 1/3. Results from the `test_memory_usage` test, with 
```
ensemble_size = 100
num_params = 1000
num_obs = 10000
```
Even bigger effect can be expected if `num_params` is larger.

tommyod:
@dafeda - yay or nay?

dafeda:
> @dafeda - yay or nay?

Good idea!
I suggest we consider this more carefully after the adaptive localization milestone.

tommyod:
[...]

PR #162, opened 2023-10-19
Facilitate streaming (row-by-row update) in ESMDA

tommyod:
To support row-by-row (parameter by parameter) computation, either literally by updating a single parameter or updating batches, we should avoid doing duplicate work.
Consider the equation:

$$C_{MD}  (C_{DD} + \alpha C_D)^{-1}  (D - Y)=X Y^T /(N-1)  (C_{DD} + \alpha C_D)^{-1}  (D - Y)$$

If we update row by row and introduce a transition matrix $K := Y^T /(N-1)  (C_{DD} + \alpha C_D)^{-1}  (D - Y) $,
then $[x_1 | x_2 | \ldots]^T K = [x_1 K | x_2 K | \ldots ]^T$. In other words, we can pre-compute $K$, since it's independent on $X$, and apply it row-by-row or group-by-group, instead of re-computing it for each group.

----------------------------

This PR proposes two methods for this:
- `compute_transition_matrix`, which computes the transition matrix $K$
- `perturb_observations`, called by both `compute_transition_matrix` and `assimilate`, to avoid duplicating code

Note: this PR assumes #161 will be merged.

-----------------------------

**I also realized that we don't have to center both $X$ and $Y$!** So I changed the code away from that, saving both memory and time by only centering the smaller matrix $Y$. See [this comment](https://github.com/equinor/iterative_ensemble_smoother/issues/160#issuecomment-1770981302). This is also documented in the code with a few lines.

### DIFF
--- a/tests/test_esmda_inversion.py
+++ b/tests/test_esmda_inversion.py
@@ -68,8 +68,8 @@ class TestEsmdaInversion:
         ans = function(alpha=alpha, C_D=C_D, D=D, Y=Y, X=X)
         K = function(alpha=alpha, C_D=C_D, D=D, Y=Y, X=None, return_K=True)
 
-        X_centered = X - np.mean(X, axis=1, keepdims=True)
-        assert np.allclose(X + ans, X + X_centered @ K)
+        X - np.mean(X, axis=1, keepdims=True)
+        assert np.allclose(X + ans, X + X @ K)
 
     @pytest.mark.parametrize("length", list(range(1, 101, 5)))
     def test_that_the_sum_of_normalize_alpha_is_one(self, length):
--- a/tests/test_esmda_inversion.py
+++ b/tests/test_esmda_inversion.py
@@ -37,6 +37,40 @@ TRUNCATION_INVERSIONS = [
 
 
 class TestEsmdaInversion:
+    # Functions that take the parameter return_K
+    @pytest.mark.parametrize(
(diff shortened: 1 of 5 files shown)


tests/test_esmda_inversion.py
```diff
@@ -37,6 +37,40 @@
...
+        # Create observations
+        D = np.random.randn(num_outputs, ensemble_members)
+        Y = np.random.randn(num_outputs, ensemble_members)
+        X = np.random.randn(num_inputs, ensemble_members)
+
+        # Test both with and without X / return_K
+        ans = function(alpha=alpha, C_D=C_D, D=D, Y=Y, X=X)
+        K = function(alpha=alpha, C_D=C_D, D=D, Y=Y, X=None, return_K=True)
+
+        X - np.mean(X, axis=1, keepdims=True)
```
tommyod, line 71: Remove this
```suggestion
        X - np.mean(X, axis=1, keepdims=True)
```

Blunde1 (approved):
LGTM.
This is very nice! In particular, see comparison between low-level and high-level API in tests. Good job!
Note for later: the names `inversion` are slightly missleading, they all involve inversion, but their purpose is either to return `K` or multiply `X` with `K`. Perhaps we can come up with good naming. As discussed, they are not part of the public api, so we may think of it later.

PR #163, opened 2023-10-19
Test that allowing ESMDA overwrites does save memory

tommyod:
Simple test :)

Also moved `test_that_float_dtypes_are_preserved` away from the `TestESMDAMemory` class, since it's not a memory test.


Blunde1 (approved):
LGTM 👍

PR #168, opened 2023-11-02
Adaptive localization

### LINKED ISSUE #167: Implement adaptive localization
See https://github.com/equinor/ert/blob/20cfcdefd1b19124a89f347ae6c8afae2d71896e/src/ert/analysis/_es_update.py#L477-L490

tommyod:
Playing around trying to come up with an API for adaptive localization.

- A new class `BaseESMDA` handles `perturb_observations` and storing covariance and observations - this is common to both:
  - The class `ESMDA`, which inherits from `BaseESMDA`, complementing it with `assimilate`.
  - The class `AdaptiveESMDA`, which also inherits from `BaseESMDA`, complementing it with its own `assimilate` method that also handles adaptive localization.

Closes #167


tommyod:
A first draft is ready: https://github.com/equinor/iterative_ensemble_smoother/blob/6a91f4c79fe328a32dc3f329eb3be1cbf40bc49a/src/iterative_ensemble_smoother/experimental.py#L139

dafeda:
The idea of rows matching the parameter group may not be great, but it was definitely not premature optimization as it made it possible to run adaptive localization on Troll in something resembling a reasonable amount of time.

dafeda:
Overall I think this looks like a really nice and clean approach.

dafeda:
I've also tried running it a bit with varying number of parameters and it seems to run amazingly fast!
Exciting! 🙀

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -6,8 +6,248 @@
...
+        assert corr_XY.max() <= 1
+        assert corr_XY.min() >= -1
+
+        # Determine which elements in the cross covariance matrix that will
+        # be set to zero
+        thres = self.correlation_threshold(ensemble_size=X.shape[1])
+        cov_XY[corr_XY < thres] = 0  # Set small values to zero
+        # print("  Entries in cov_XY set to zero", np.isclose(cov_XY, 0).sum())
+
+        return X + cov_XY @ transition_matrix
```
Blunde1, line 125: This works.
Perhaps intuitively for others, because sample covariance among points does not include other variables.
Not intuitively for me, as I know that we may decompose our objective to `cov_x @ LLS_coeff @ np.linalng.inv(cov_d)`. The `LLS_coeff` _should_ and does indeed change when removing variables. But this is captured by _sample covariance_ `cov_x` in the above, so the effect is removed.
So the above works.

Blunde1: I must admit that I do not like this (mathematical) property.
But I do like the algorithmic exploitation of it in the code. 👍

Blunde1: > This works. 

Here referring to that we do not have to recompute `cov_XY` after variable removal / set to zero.

Blunde1:
> However, when the threshold is between `0` and `1`, I get completely nonsensical results. Here with `0.5`: image

The problem seems to worsen at increased number of ES-MDA iterations, or at least depend heavily upon it. Here is a plot running with only iteration (`alpha = normalize_alpha(np.ones(1))`) should correspond to ES with adaptive localization. The results are still poor, so it should be possible to debug this (simpler) case.

image

Blunde1:
@kvashchuka, @dafeda and me had a whiteboard session (final result after some iterations is included).

The hypothesis was to inspect the "transition matrix" in the code, let's call it $T$, which should be an estimator of $\Sigma_d^{-1}(d_i-y_i)$, note that we have multiple forms of the observation covariance $\Sigma_d\approx (H\hat{\Sigma_x}H^T + \Sigma_{\epsilon}) \approx (YY^T/(n-1) + \Sigma_{\epsilon}) $, where the latter is employed in ES.

What we think the problem, is that $T$ is calculated once with the full sets of observations and responses. For global $T$ to be the "correct" transition matrix for a parameter $i$, say $T=T_i$, we would require our estimate of $\Sigma_d$ to be diagonal. This is never true for $T$ using the sample covariance $YY^T/(n-1)$ internally, making the estimate $\hat{\Sigma_d}$ dense, and generally therefore $T\neq T_i$.

The original implementation of adaptive localization implicitly found a $T_i$ for each parameter $i$, by subsetting the $Y$'s and $D$'s which were deemed "significant" to affect the posterior update of parameter $i$.
This was computationally expensive, because finding $T_i$ is computationally expensive through inversion, and we have a great deal of them due to the many parameters.

The batching algorithm sought subsets of $i$'s for which $T_i$ should be the same. The search is non-random because consecutive parameters in $X$ are "likely" close in space if they correspond to some spatial parameters in a discretized field. This search is done before inversion and finding $T_{subset}$.

In statistical terms, the marginal covariance is indeed the subsetted covariance directly, however this does not hold generally true for the inverse (precision) which we are working with. It holds true if the precision is diagonal, or if it is block-diagonal and we are subsetting the blocks. This should be equivalent to finding block $T_{subset}$'s. Perhaps we could use the $\Sigma_{x,y}$ and cross-correlations/thresholding to be even smarter about this search?

image

tommyod:
Not sure I follow. You first use $i$ as a reference to the transition matrix and the observation vector $y_i$, then say "transition matrix for a parameter $i$", but the transition matrix is independent of the parameters. Is $i$ a parameter or observation index?

---------------------------

The update equation for all parameters is:

$X^{\text{updated}} = X + XY^T / (n-1) \Sigma_d^{-d} (d - y) = X + \text{cov}(X, Y)T$

For a single parameter $i$, a row vector $x_i$ in $X$, the update equation becomes:

$x_i^{\text{updated}} = x_i + x_iY^T / (n-1) \Sigma_d^{-d} (d - y) = x_i + \text{cov}(x_i, Y)T$

So $T$ is completely independent of which $i$ we're updating. So it seems to me that $T$ is always the correct matrix, no matter what $i$ we're updating.

dafeda:
> * You first use i as a reference to the transition matrix and the observation vector yi, then say "transition matrix for a parameter i", but the transition matrix is independent of the parameters. Is i a parameter or observation index?

The transition matrix depends on the parameters via the Y = g(X).

> ```
> x_i_updated = x_i + cov(x_i, Y[m, :]) inv(cov(Y[m, :], Y[m, :]) + sigma) @ (d -Y)[m, :]
> ```
> 
> I agree that this is something different, since `inv(cov(Y[m, :], Y[m, :]) != inv(cov(Y, Y)[m, :]`

Indeed 👍

tommyod:
I implemented a version with the masking. Results look much much better:
image

Blunde1:
One of the great things about the KF is that it runs on $O(m)$ time for updating instead of $O(m^3)$ when we have $m$ measurements. This is possible when the measurement error is independent, and consequently the covariance of the error is diagonal. See Understanding the EnKF Section 4.2 Serial Updating. https://folk.ntnu.no/joeid/Katzfuss.pdf it refers to this paper that I believe is the origin paper for covariance localization https://www.researchgate.net/publication/228817876_A_Sequential_Ensemble_Kalman_Filter_for_Atmospheric_Data_Assimilation

Our measurement error is so independent so that we have even built it into the api I believe. At least for current practical purposes, $\Sigma_\epsilon$ will be diagonal. Then why can we too not do sequential updates? There probably is a good reason and I am not seeing it.

Thus throwing something wild out there: how about iterating over observations assimilating them sequentially, and then iterate over parameters when doing this sequential update?
```python
for (dj, yj, Sigma_eps_j) in (observations, responses, variance):
    H = LLS coefficients yj on X[mask,:] # 1xp but using X which could be p2xn for p2<p depending on mask
    Sigma_yj = H @X @ X.T @ H.T # 1x1
    Sigma_d = Sigma_yj + Sigma_eps_j # 1x1
    for i in realizations
        T_ji = (dj - yji) / Sigma_d # 1x1
        for X_ki in realization_i_of_parameters:
            if dj not masked for xk:
                X_ki = X_ki + cov_xk_yj*T_ji
```

Is this easily wrong? Or wrong for some other reason? I tried to adjust for the weird reasons that using the full transition matrix was wrong (writing `H = LLS coefficients yj on X[mask:,]`). It therefore assumes the masking is pre-computed. Is this unreasonable?

tommyod:
Sketch...

------------------

I'll use the ESMDA notation from the Emerick paper, copying from a previous issue. The update is

$$X_\text{updated} = C_{MD}  (C_{DD} + \alpha C_D)^{-1}  (D - Y),$$

and the linear model is $g(X) = G X$, where $G$ is a matrix. Then $C_{DD} = \text{cov}(Y, Y) = G \text{cov}(X, X) G^T$ if the function $g$ is a linear map. Furthermore $C_{MD} = \text{cov}(X, Y) = \text{cov}(X, X) G^T$ if the function $g$ is a linear map.

$$X_\text{updated} = \text{cov}(X, X) G^T  (G \text{cov}(X, X) G^T + \alpha C_D)^{-1}  (D - Y)$$

------------------

```python3
# I assume Y has shape (num_obs, num_ensemble)
for (dj, yj, Sigma_eps_j) in (observations, responses, variance):
    # I assume y_j will have shape (1, num_ensemble)

    # If X has shape (num_params, num_ensemble), then
    # then then relevant coefficients of H must have shape (1, num_params)
    # for the equation yj = H X to hold.
    H = LLS coefficients yj on X[mask,:] # Shape (1, num_params)

    # The shape of (H @ X) becomes (1, num_ensemble)
    Sigma_yj = H @X @ X.T @ H.T # Shape does become (1, 1)
    Sigma_d = Sigma_yj + Sigma_eps_j # 1x1
    for i in realizations
        T_ji = (dj - yji) / Sigma_d # 1x1
        for X_ki in realization_i_of_parameters:
            if dj not masked for xk:
                X_ki = X_ki + cov_xk_yj*T_ji
```

Blunde1:
> Pretty sure we discussed this a long time ago and agreed that it made sense.

That would be really cool! In case you also thinks it holds I suggest we code and test it on the new [notebook example](https://github.com/equinor/iterative_ensemble_smoother/blob/main/docs/source/LinearRegression.py)

tommyod:
We're drifting off topic. This PR is about implementing a decent API and a working algorithm for adaptive localization. Further investigations into speedsup and statistics will definitely be part of this, but let's do it one PR/issue at at time.

I have opened the following issue for continuation of this specific discussion: https://github.com/equinor/iterative_ensemble_smoother/issues/175

tommyod:
**Ready for a review.** :+1: 

In summary I ended up with:
- A base class `BaseESMDA` that is used as `ESMDA(BaseESMDA)` and `AdaptiveESMDA(BaseESMDA)`. Some functionality is common to both algorithms. This refactoring is backward-compatible wrt `ESMDA`.
- A new class `AdaptiveESMDA` for adaptive localization.
- Test for cutoffs `1` and `0`, plus the test `test_that_posterior_generalized_variance_increases_in_cutoff`. This was implemented based on tests by @kvashchuka  and @Blunde1 .
- A test `test_that_adaptive_localization_works_with_dying_realizations` that is part example, part test. It shows the full computation with (1) dying realizations due to nodes going down and (2) updating parameter groups (streaming). 

There is still some work to do, represented as TODOs in the code. Also, we should time the code and look into performance wrt both time and memory. I want to do this in future, smaller PRs however, since I am reasonably happy with the overall API design now.

tests/test_esmda.py
```diff
@@ -25,9 +25,55 @@
...
+        Y = rng.normal(size=(num_outputs, num_ensemble))
+        covariance = np.exp(rng.normal(size=num_outputs))
+        observations = rng.normal(size=num_outputs, loc=1)
+
+        # Create ESMDA instance and perform one iteration
+        esmda = ESMDA(
+            covariance, observations, alpha=alpha, seed=seed + 99, inversion="exact"
+        )
+        X_ESMDA = np.copy(X)
+        for _ in range(esmda.num_assimilations()):
```
dafeda, line 60: Do we need this for-loop or is it enough to just call `assimilate` once?

src/iterative_ensemble_smoother/esmda.py
```diff
@@ -34,9 +35,100 @@
...
+    ) -> npt.NDArray[np.double]:
+        """Create a matrix D with perturbed observations.
+
+        In the Emerick (2013) paper, the matrix D is defined in section 6.
+        See section 2(b) of the ES-MDA algorithm in the paper.
+
+        Parameters
+        ----------
+        ensemble_size : int
+            The ensemble size, i.e., the number of perturbed observations.
```
dafeda, line 101: Not sure we need the "i.e., the number of perturbed observations"

src/iterative_ensemble_smoother/esmda.py
```diff
@@ -92,35 +184,16 @@ def __init__(
...
-
-        if covariance.ndim == 2 and covariance.shape[0] != covariance.shape[1]:
-            raise ValueError("Argument `covariance` must be square if it's 2D.")
 
-        if not (isinstance(observations, np.ndarray) and observations.ndim == 1):
-            raise TypeError("Argument `observations` must be a 1D NumPy array.")
-
-        if not observations.shape[0] == covariance.shape[0]:
-            raise ValueError("Shapes of `observations` and `covariance` must match.")
+        # Initialize the super-class
```
dafeda, line 188: Comment not needed in my opinion.

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -6,8 +6,241 @@
...
+            cov_XY @ inv(C_DD + alpha * C_D) @ (D - Y)
+            center(X) @ center(Y) / (N - 1) @ inv(C_DD + alpha * C_D) @ (D - Y)
+            center(X) @ [center(Y) / (N - 1) @ inv(C_DD + alpha * C_D) @ (D - Y)]
+
+        The advantage of forming T instead is that if we already have computed
+        cov_XY, then we can apply the same T to a reduced number of rows (parameters)
+        in cov_XY.
+        """
+
+        # TODO: This is re-computed in each call. Can we pre compute it?
```
dafeda, line 111: I think the current implementation does this outside the loop over parameter groups:

https://github.com/equinor/ert/blob/6aa2aaf5d2a0932c740f95ca1143a98c6460f620/src/ert/analysis/_es_update.py#L544-L550

Does this mean that we would recompute this for ever batch of parameters?

tests/test_esmda.py
tommyod: in this particular test it's not needed. I can remove it :)

src/iterative_ensemble_smoother/experimental.py
tommyod: Currently yes. I will look into fixing this before merging! Should be an easy fix.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+    X = rng.normal(size=(num_parameters, num_ensemble))
+    covariance = rng.triangular(0.1, 1, 1, size=num_observations)
+    yield X, g, observations, covariance, rng
+
+
+class TestAdaptiveESMDA:
+    @pytest.mark.parametrize(
+        "linear_problem",
+        range(25),
+        indirect=True,
```
dafeda, line 80: I have not used `indirect=True` before so bear with me.
The way I read this is that we run the test 25 times where before each run `linear_problem` is run with `rng` set as `default_rng(i)`.
What's the benefit of this versus running once with `rng` set to `default_rng()` so that we get a different seed every time the test is run?
I guess we get broader coverage since the test is run with 25 different seeds, but over time using `default_rng()` will have even better coverage since a new seed is used every time.
I may of course just be misunderstanding how this works...

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+        ids=[f"seed-{i+1}" for i in range(25)],
+    )
+    def test_that_adaptive_localization_with_cutoff_1_equals_ensemble_prior(
+        self, linear_problem
+    ):
+
+        # Create a problem with g(x) = A @ x
+        X, g, observations, covariance, rng = linear_problem
```
dafeda, line 88: `rng` is not used in this test.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+
+        # Create adaptive smoother
+        smoother = AdaptiveESMDA(
+            covariance=covariance, observations=observations, seed=1
+        )
+
+        X_i = np.copy(X)
+        alpha = normalize_alpha(np.ones(5))
+        for i, alpha_i in enumerate(alpha, 1):
```
dafeda, line 97: `i` is not used so no need for `enumerate`.

tests/test_experimental.py
dafeda: As discussed on Slack, I agree that this is a good thing to do but that using `hypothesis` might be a better approach. We do not have to make the switch in this PR though.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+                ensemble_size=Y_i.shape[1], alpha=alpha_i
+            )
+
+            # Update the relevant parameters and write to X (storage)
+            X_i = smoother.assimilate(
+                X=X_i,
+                Y=Y_i,
+                D=D_i,
+                alpha=alpha_i,
+                correlation_threshold=lambda ensemble_size: 1,
```
dafeda, line 113: Would it be good if we could set `correlation_threshold` as a `float` as well as a `Callable`?
Seems kind of natural to call `assimilate` with `correlation_threshold=1.0`.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+
+            # Run forward model
+            Y_i = g(X_i)
+
+            # Create noise D - common to this ESMDA update
+            D_i = smoother.perturb_observations(
+                ensemble_size=Y_i.shape[1], alpha=alpha_i
+            )
+
+            # Update the relevant parameters and write to X (storage)
```
dafeda, line 107: Think we can remove the reference to `storage` here as it's `ert` specific.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+        indirect=True,
+        ids=[f"seed-{i+1}" for i in range(25)],
+    )
+    @pytest.mark.parametrize(
+        "cutoffs", [(0, 1e-3), (0.1, 0.2), (0.5, 0.5 + 1e-12), (0.9, 1), (1 - 1e-3, 1)]
+    )
+    def test_that_posterior_generalized_variance_increases_in_cutoff(
+        self, linear_problem, cutoffs
+    ):
+        """As the number of ensemble members decrease, this test starts to fail
```
dafeda, line 188: Should this be `As the number of ensemble members **increases**`?

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+
+            # Create noise D - common to this ESMDA update
+            D_i = smoother.perturb_observations(
+                ensemble_size=Y_i.shape[1], alpha=alpha_i
+            )
+
+            cutoff_low, cutoff_high = cutoffs
+            assert cutoff_low <= cutoff_high
+
+            # Update twice, one with low cutoff, once with low cutoff
```
dafeda, line 218: Typo: "once with low cutoff, once with high cutoff".
I vote for removing the comment.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+            X_i_high_cutoff = smoother.assimilate(
+                X=X_i,
+                Y=Y_i,
+                D=D_i,
+                alpha=alpha_i,
+                correlation_threshold=lambda ensemble_size: cutoff_high,
+            )
+
+            # Compute covariances
+            prior_cov = np.cov(X, rowvar=False)
```
dafeda, line 235: Should `rowvar` be `True` here? I would expect `prior_cov` to have dimension `(num_parameters, num_parameters)` and not `(num_ensemble, num_ensemble)`.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+            posterior_cutoff_low_cov = np.cov(X_i_low_cutoff, rowvar=False)
+            posterior_cutoff_high_cov = np.cov(X_i_high_cutoff, rowvar=False)
+
+            # Compute determinants of covariance matrices
+            # https://en.wikipedia.org/wiki/Generalized_variance
+            # intuitively: large determintant => high covariancce
+            #  => smaller volume of multivariate normal
```
dafeda, line 242: My intuition is that a large determinant means more spread and thus a larger volume. Am I wrong?

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+        X, g, observations, covariance, rng = linear_problem
+        num_parameters, num_ensemble = X.shape
+        num_observations = len(observations)
+        rng = np.random.default_rng()
+
+        # Split the parameters into two groups of equal size
+        num_groups = 10
+        assert num_observations % num_groups == 0, "Num groups must divide parameters"
+        group_size = num_parameters // num_groups
+        parameters_groups = list(zip(*(iter(range(num_parameters)),) * group_size))
```
dafeda, line 292: I find this a bit hard to follow.
Could we use `np.array_split` instead? Something like this:

`np.array_split(np.arange(num_parameters), num_groups)`

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+        assert len(parameters_groups) == num_groups
+
+        # =============================================================================
+        # SETUP ESMDA FOR LOCALIZATION AND SOLVE PROBLEM
+        # =============================================================================
+        alpha = normalize_alpha(
+            np.array([5, 4, 3, 2, 1])
+        )  # Vector of inflaction values
```
dafeda, line 300: Typo: should be `inflation`.

tests/test_experimental.py
```diff
@@ -1,14 +1,392 @@
...
+        start_time = time.perf_counter()
+        smoother = AdaptiveESMDA(
+            covariance=covariance, observations=observations, seed=1
+        )
+
+        # Simulate realization that die
```
dafeda, line 306: Typo: should be `realizations`.

tests/test_experimental.py
kvashchuka: No, the flakiness in the same test in ert was due to too small ensemble size. Also, the math behind it tells us to expect this behaviour when ensmeble_size -> inf

dafeda: """As the number of ensemble members decrease, this test starts to fail
more often. The property only holds in the limit as the number of
ensemble members goes to infinity."""

I am probably acting as a badly trained `GPT` but this docstring confused me as I thought the second sentence referenced the first sentence.

src/iterative_ensemble_smoother/esmda.py
```diff
@@ -34,9 +35,96 @@
...
+            Each column consists of perturbed observations, scaled by alpha.
+
+        """
+        # Draw samples from zero-centered multivariate normal with cov=alpha * C_D,
+        # and add them to the observations. Notice that
+        # if C_D = L @ L.T by the cholesky factorization, then drawing y from
+        # a zero cented normal means that y := L @ z, where z ~ norm(0, 1).
+        # Therefore, scaling C_D by alpha is equivalent to scaling L with sqrt(alpha).
+
+        D: npt.NDArray[np.double] = self.observations[:, None] + np.sqrt(
```
dafeda, line 120: We are using `np.newaxis` in other places in the code so I suggest we stick to that.
Also, this allows me to try out the suggestion-feature in GitHub 😄 

```suggestion
        D: npt.NDArray[np.double] = self.observations[:, np.newaxis] + np.sqrt(
```

tests/test_experimental.py
tommyod: I've added this, but I am on the fence. It's easy enough to add, so I did it.

tests/test_experimental.py
```diff
@@ -1,14 +1,388 @@
...
+from iterative_ensemble_smoother.esmda_inversion import normalize_alpha
 from iterative_ensemble_smoother.experimental import (
+    AdaptiveESMDA,
     ensemble_smoother_update_step_row_scaling,
+    groupby_indices,
 )
 
 
+@pytest.mark.parametrize("seed", range(25))
+def test_groupby_indices(seed):
```
dafeda, line 18: From my reading of this test, it does not test the "grouping" property of the function when multiple rows of `X` are identical.
Since we assert that `union_idx == set(range(rows))` we I think assert that `X` has no duplicate rows.
Am I understanding this correctly?

dafeda (approved):
This is great 👍

PR #171, opened 2023-11-16
Notebook with linear regression (Gauss-Linear model)

tommyod:
- Added the Gauss-Linear case in a notebook.
- I believe this is a nice notebook to experiment with, since the Gauss-Linear case is what ESMDA is based on.
- The notebook allows the user to easy experiment with (1) number of ensemble members, (2) number of parameters, (3) number of observations and (4) strength of the prior.

image

### DIFF
--- a/docs/source/LinearRegression.py
+++ b/docs/source/LinearRegression.py
@@ -1,0 +1,172 @@
+# ---
+# jupyter:
+#   jupytext:
+#     formats: ipynb,py:percent
+#     text_representation:
+#       extension: .py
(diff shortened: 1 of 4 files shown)


Blunde1 (approved):
I read through the notebook, and it looks brilliant. Great job 🚀 
I had one small wish, but I can write that in an issue for later:
- if we sample from a prior more aligned with the linear model, we should expect better posterior results than the uninformed maximum likelihood, right?

Anyway, probably good to just get this in ASAP without adding anything.

PR #172, opened 2023-11-16
Update README installation instructions

tommyod:


kvashchuka (approved):
Looks good! I suggest to squash three commits into one and it is good to go :)

PR #173, opened 2023-11-16
Added LITERATURE.md

tommyod:
- I thought maintaining a `LITERATURE.md` Markdown might be easier than adding it to sphinx.
- Also, the sphinx docs should arguably only documen what we have actually implemented, but I want this documentation list to be more general, having e.g. preliminary readings, things we are interesting in, but have not implemented, etc.

After merging others can add their sources too. I don't want to add everything for everyone, since I don't have the full overview.

### DIFF
--- a/LITERATURE.md
+++ b/LITERATURE.md
@@ -1,0 +1,30 @@
+# Literature
+
+A list of literature that we find useful for Data Assimilation.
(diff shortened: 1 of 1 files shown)


dafeda (approved):
Nice list 👍

Issue #180, opened 2023-11-23
Use of `K` in docstring of `inversion_exact_cholesky`


dafeda:
https://github.com/equinor/iterative_ensemble_smoother/blob/1cc21ee6e3979fc1c85e4c0a287a9102a16d428a/src/iterative_ensemble_smoother/esmda_inversion.py#L193-L202

Docstring defined `K = inv(C_DD + alpha * C_D) @ (D - Y)` but when I see `K` I see Kalman gain so I suggest we try and find a different letter.

PR #181, opened 2023-11-23
Use T for transition matrix instead of K

dafeda:
K usually refers to Kalman gain

Resolves: #180


tommyod (approved):
Very nice. Both the errors in docs, and the change in notation. I agree that a T is better name.

PR #188, opened 2023-11-29
Adaptive Localization notebook

tommyod:
I created a notebook to understand better how Adaptive Localization works. The notebook contains the simplest problem I could think of where I expect adaptive localization to outperform standard ESMDA -- a Gauss-Linear problem with sparse structure in the forward model. I hope this notebook and problem can be used for quality assurance in the future if we e.g. experiment with thresholding, add the Lasso idea, etc.

A Gauss-linear problem with a very local structure in the forward model. Each output is determined by at most 3 parameters. The forward map has this structure:

image

## Results (correlation coefficients)

Some results with the default threshold $3 / \sqrt{N}$

### 50 parameters, 50 observations

image

### 100 parameters, 50 observations

image

### 500 parameters, 50 observations

image

## Comments on results

image

- First of all, I am happy to see that in general, the adaptive localization seems to work. It works especially well with many ensemble members.
- When `num_ensemble` <= 9, the AdaptiveESMDA posterior equals the prior. No information is used. This happens irrespective of the problem size. If you have a lot of parameters this might make sense, but consider a small-parameter case (1 parameter and 1 response being the extreme example) - it does not make sense much sense to me to not update a prior if you have 9 observations and you're trying to estimate one parameter.
- On this problem, it seems like `num_ensemble` has to surpass ~15 for adaptive localization to correlate better with true parameter values compared to ESMDA.

### DIFF
--- a/docs/source/Adaptive.py
+++ b/docs/source/Adaptive.py
@@ -88,6 +88,26 @@ plt.tight_layout()
 plt.show()
 
 
+# %% [markdown]
+# - Below we draw prior realizations $X \sim N(0, \sigma)$.
+# - The true parameter values used to generate observations are in the range $[-1, 1]$.
+# - As the number of realizations (ensemble members) goes to infinity,
+#   the correlation between the prior and the true parameter values converges to zero.
+# - The correlation is zero for a finite number of realizations too,
+#   but statistical noise might induce some spurious correlations between
+#   the prior and the true parameter values.
+#
+# **In summary the baseline correlation is zero.**
+# Anything we can do to increase the correlation above zero beats the baseline,
+# which is using the prior (no update).
+#
+# The correlation coefficient does not take into account the uncertainty
+# represeted in the posterior, only the mean posterior value is compared with
+# the true parameter values.
+# To compare distributions we could use the
+# [Kullback–Leibler divergence](https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence),
+# but we do not pursue this here.
+
 # %%
 def get_prior(num_parameters, num_ensemble, prior_std):
     """Sample prior from N(0, prior_std)."""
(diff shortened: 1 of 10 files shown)


docs/source/Adaptive.py
```diff
@@ -0,0 +1,333 @@
...
+# Initial ensemble X ~ N(0, prior_std) and diagonal covariance with ones
+X = rng.normal(size=(num_parameters, num_ensemble)) * prior_std
+
+# Covariance matches the noise added to observations above
+covariance = np.ones(num_observations)  # N(0, 1) covariance
+
+# %% [markdown]
+# ## Solve the maximum likelihood problem
+#
+# We can solve $Ax = b$, where $b$ is the observations,
```
dafeda, line 109:
```suggestion
# We can solve $Ax = b$, where $b$ are the observations,
```

docs/source/Adaptive.py
```diff
@@ -0,0 +1,333 @@
...
+plt.xlabel("Parameter index")
+plt.ylabel("Parameter value")
+plt.grid(True, ls="--", zorder=0, alpha=0.33)
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# ## Solve using ESMDA
+#
+# We crease an `ESMDA` instance and solve the Guass-linear problem.
```
dafeda, line 130:
```suggestion
# We create an `ESMDA` instance and solve the Guass-linear problem.
```

docs/source/Adaptive.py
```diff
@@ -0,0 +1,333 @@
...
+plt.xlabel("Parameter index")
+plt.ylabel("Parameter value")
+plt.grid(True, ls="--", zorder=0, alpha=0.33)
+plt.legend()
+plt.show()
+
+# %% [markdown]
+# ## Solve using AdaptiveESMDA
+#
+# We crease an `AdaptiveESMDA` instance and solve the Guass-linear problem.
```
dafeda, line 173:
```suggestion
# We create an `AdaptiveESMDA` instance and solve the Guass-linear problem.
```

tommyod:
Added the prior too:
image

Blunde1 (approved):
Very nice! Comment on natural baseline on slack. Approved, LGTM.

PR #193, opened 2023-12-06
Profiling AdaptiveESMDA

### LINKED ISSUE #192: Line profile AdaptiveESMDA
- Run line profiling
- Figure out where time is spent
- See if we can speed it up

tommyod:
Closes #192

## AdaptiveESMDA is slower in general

Some rough measurements first. On the sparse Gauss-Linear problem with size
```python
num_parameters = 10_000
num_observations = 1000
num_ensemble = 25
```
ESMDA takes `234 ms` while AdaptiveESMDA takes `4.13 s`.

Going up to
```python
num_parameters = 25_000
```
ESMDA takes `308 ms` while AdaptiveESMDA takes `9.7 s`.

Going up to
```python
num_parameters = 50_000
```
ESMDA takes `539 ms` while AdaptiveESMDA takes `19 s`.

-----------------------------

## Profiling `AdaptiveESMDA.assimilate`

Here is the report from 
```
%lprun -f AdaptiveESMDA.assimilate adaptive_smoother.assimilate(X_i, Y=Y_i, D=D_i, alpha=alpha_i, verbose=False)
```
the sparse Gauss-Linear problem with size
```python
num_parameters = 10_000
num_observations = 1000
num_ensemble = 25
```
Only the important parts are pasted here:

```
Timer unit: 1e-09 s

Total time: 0.795092 s
File: iterative_ensemble_smoother/src/iterative_ensemble_smoother/experimental.py
Function: assimilate at line 210

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   210                                               def assimilate(self, X, Y, D, alpha, correlation_threshold=None, verbose=False):
   211                                                   """Assimilate data and return an 
   287         1   66322453.0    7e+07      8.3          cov_XY = empirical_cross_covariance(X, Y)
   288         1       1266.0   1266.0      0.0          assert cov_XY.shape == (X.shape[0], Y.shape[0])
   289         1   82010253.0    8e+07     10.3          corr_XY = self._correlation_matrix(cov_XY, X, Y)
   290         1       4196.0   4196.0      0.0          assert corr_XY.shape == cov_XY.shape
   291                                           
   292                                                   # Determine which elements in the cross covariance matrix that will
   293                                                   # be set to zero
   294         1      26608.0  26608.0      0.0          threshold = correlation_threshold(ensemble_size=X.shape[1])
   295         1   17874591.0    2e+07      2.2          significant_corr_XY = np.abs(corr_XY) > threshold
   296                                           
   297                                                   # Pre-compute the covariance cov(Y, Y) here, and index on it later
   298         1    6570085.0    7e+06      0.8          cov_YY = empirical_cross_covariance(Y, Y)
   302      2166  414649781.0 191435.7     52.2          for (unique_row, indices_of_row) in groupby_indices(significant_corr_XY):
   303                                           
   332                                                       # Compute transition matrix T
   333      4328   99213275.0  22923.6     12.5              T = self.compute_cross_covariance_multiplier(
   338      2164     247438.0    114.3      0.0                  cov_YY=cov_YY_subset,  # Passing cov(Y, Y) avoids re-computation
   339                                                       )
   340      2164   14599430.0   6746.5      1.8              X_out[indices_of_row, :] = X_subset + cov_XY_subset @ T
```

- Computing `cov_XY` and `corr_XY` is expensive, since it takes a product of two large matrices. This is not done explicitly in ESMDA because of the order of matrix multiplication.
- `groupby_indices` is also costly.
- `compute_cross_covariance_multiplier` takes some time too.

-----------------------------

## Examining `groupby_indices`

My implementation of `groupby_indices` was based on [Fedas implementation](https://github.com/equinor/ert/blob/399d966e35f20c071b0cf82987ed423fa459bb66/src/ert/analysis/_es_update.py#L588).
I never actually compared the speed, so let's do that.

```python
def groupby_indices_feda(X):
    """Minimal implementation of groupby_indices."""
    param_correlation_sets = np.unique(X, axis=0)
    row_with_all_false = np.all(~param_correlation_sets, axis=1)
    param_correlation_sets = param_correlation_sets[~row_with_all_false]

    for param_correlation_set in param_correlation_sets:
        matching_rows = np.all(X == param_correlation_set, axis=1)
        row_indices = np.where(matching_rows)[0]
        yield (param_correlation_set, row_indices)
```

The new implementation seems to always be faster. How much faster it is depends on the structure of the matrix. The more parameter groups, the larger the difference in speed.

```
X = np.repeat((np.random.randn(50, 500) > 0.5), 50, axis=0)

X.shape # (2500, 500)

%timeit list(groupby_indices_feda(X))
110 ms ± 216 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit list(groupby_indices(X))
46.4 ms ± 216 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

X = np.repeat((np.random.randn(200, 500) > 0.5), 200, axis=0)

%timeit list(groupby_indices_feda(X))
3.56 s ± 25.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit list(groupby_indices(X))
807 ms ± 8.73 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

X = np.random.randn(100 * 100, 500) > 0.5

%timeit list(groupby_indices_feda(X))
8.01 s ± 494 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit list(groupby_indices(X))
113 ms ± 2.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

It seems like the current implemenation is faster than the initial one. It also seems unlikely that this can be sped up, since it's more or less linear.

### DIFF
--- a/docs/source/Adaptive.py
+++ b/docs/source/Adaptive.py
@@ -1,0 +1,378 @@
+# ---
+# jupyter:
+#   jupytext:
+#     formats: ipynb,py:percent
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.15.2
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %%
+# ruff: noqa: E402
+
+# %% [markdown]
+# # Adaptive Localization
+#
+# In this example we run adaptive localization
+# on a linear sparse problem.
+#
+# - Each response is only affected by $3$ parameters.
+# - This is represented by a tridiagonal matrix $A$ in the forward model $g(x) = Ax$.
+# - The problem is Gauss-Linear, so in this case ESMDA will sample
+#   the true posterior when the number of ensemble members (realizations) is large.
+
+# %% [markdown]
+# ## Import packages
+
+# %%
+import numpy as np
+import scipy as sp
+from matplotlib import pyplot as plt
+
+from iterative_ensemble_smoother import ESMDA
+from iterative_ensemble_smoother.experimental import AdaptiveESMDA
+
+# %% [markdown]
+# ## Create problem data
+#
+# Some settings worth experimenting with:
+#
+# - Decreasing `prior_std=1` will pull the posterior solution toward zero.
+# - Increasing `num_ensemble` will increase the quality of the solution.
+# - Increasing `num_observations / num_parameters`
+#   will increase the quality of the solution.
+
+# %%
+rng = np.random.default_rng(42)
+
+# Dimensionality of the problem
+num_parameters = 10_000
+num_observations = 1000
+num_ensemble = 25
+prior_std = 1
+
+# Number of iterations to use in ESMDA
+alpha = 5
+
+# %% [markdown]
+# ## Create problem data - sparse tridiagonal matrix $A$
+
+# %%
+diagonal = np.ones(min(num_parameters, num_observations))
+
+# Create a tridiagonal matrix (easiest with scipy)
+A = sp.sparse.diags(
+    [diagonal, diagonal, diagonal],
+    offsets=[-1, 0, 1],
+    shape=(num_observations, num_parameters),
+    dtype=float,
+).toarray()
+
+# We add some noise that is insignificant compared to the
+# actual local structure in the forward model
+A = A + rng.standard_normal(size=A.shape) * 0.01
+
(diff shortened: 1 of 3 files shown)


tommyod:
Closing this PR and associated issue. 

Within the context of how we perform out computations, I see no obvious way to speed up computations.

Looking over the numbers might be interesting to @dafeda  and @Blunde1

PR #194, opened 2023-12-06
Fix bug when using full covariance matrix in AdaptiveESMDA

tommyod:
- Slicing for a full covariance matrix was wrong. Need to slice both dimensions.
- Added a test for this by parameterizing another test.


Blunde1 (approved):
LGTM

PR #197, opened 2023-12-13
Update README.md with how to make release

tommyod:
Might be nice to have? Saves some time searching the web for "how to create a tag" perhaps :)


dafeda (approved):
Not sure how useful this is, but let's give it a go.

PR #202, opened 2023-12-27
Add more papers to LITERATURE.md

tommyod:
Added every paper that I know of that I find useful.

Some of these are not worth reading in great detail, but they are useful to know about. So I'd rather add too many than too few. I highlighted the ESMDA and SIES papers, since they are the most important.

**Reviewers:** Not much to review. Let me know if you want me to keep the list shorter. Would be nice if others could add their papers too, either as suggestions I can approve here, or in a different PR.


LITERATURE.md
```diff
@@ -5,24 +5,50 @@ Could be moved to `docs/source` eventually.
 
 ## Papers
 
-### SIES
+### SIES (Subspace Iterative Ensemble Smoother)
 
-- (2019) [Efficient Implementation of an Iterative Ensemble Smoother for Data Assimilation and Reservoir History Matching](https://www.frontiersin.org/articles/10.3389/fams.2019.00047/full) by Evensen et al
+An optimization approach.
+The observations are perturbed once, and we use the Gauss-Newton algorithm to minimize a cost function.
+The proposed algorithm has state across iterations in a matrix $W$, which makes it more challening to reason about compared to ESMDA.
```
dafeda, line 12:
```suggestion
The proposed algorithm has state across iterations in a matrix $W$, which makes it more challenging to reason about compared to ESMDA.
```

dafeda (approved):
Nice 👍

PR #203, opened 2023-12-27
Add LASSO without structure to LinearRegression notebook

Blunde1:


docs/source/LinearRegression.py
```diff
@@ -170,3 +171,43 @@ def g(X):
...
 plt.show()
+
+# %%
+import scipy.sparse as sp
+from graphite_maps.linear_regression import linear_l1_regression
+
+# %%
+X_prior = np.copy(X)
+Y = g(X_prior)
+D = Y + rng.standard_normal(size=Y.shape)
```
tommyod, line 182: This line should use `covariance` or `obs_sd`. It works now, but only because `obs_sd=1.0`.

Blunde1:
The docs build is failing, I guess because I import graphite-maps? Will copy the imported `linear_l1_regression` function so that we avoid the import.

tommyod:
Not sure why it fails now. I see no error message... Try running locally

```
sphinx-build -c docs/source/ -b html docs/source/ docs/build/html -W
```

Blunde1:
> Not sure why it fails now. I see no error message... Try running locally
> 
> ```
> sphinx-build -c docs/source/ -b html docs/source/ docs/build/html -W
> ```

I see I need Java for the docs? (pandoc) Is there no way around this?

From checking the CI I see

```
Warning, treated as error:
/home/runner/work/iterative_ensemble_smoother/iterative_ensemble_smoother/docs/source/LinearRegression.ipynb:Executing notebook failed: CellExecutionError [mystnb.exec]
```
perhaps I import or use something in the notebook that is not part of the `pip install .[doc]` requirements?

tommyod (commented):
Added some comments!

docs/source/LinearRegression.py
```diff
@@ -170,3 +174,131 @@ def g(X):
...
+
+    return H
+
+
+# %%
+# Learn Kalman gain
+X_prior = np.copy(X)
+Y = g(X_prior)
+D = Y + obs_sd * rng.standard_normal(size=Y.shape)
+K = linear_l1_regression(D=D.T, X=X_prior.T)
```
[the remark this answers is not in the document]
Blunde1: I do believe this is correct. Passing `D` allows the regression to learn the covariance of `d`, not `y`, and it is indeed `cov(d,d)` that is used for the Kalman gain, which is estimated and extracted directly from the regression coefficients here.

docs/source/LinearRegression.py
```diff
@@ -170,3 +174,131 @@ def g(X):
...
+    n_y, m = X.shape  # m: number of y responses
+
+    # Assert that the first dimension of U and Y are the same
+    assert n == n_y, "Number of samples in D and X must be the same"
+
+    scaler_u = StandardScaler()
+    U_scaled = scaler_u.fit_transform(D)
+
+    scaler_y = StandardScaler()
+    Y_scaled = scaler_y.fit_transform(X)
```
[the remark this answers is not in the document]
Blunde1: Agreed. This is a miss in re-naming variables from the graphite-maps repo.

docs/source/LinearRegression.py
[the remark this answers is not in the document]
Blunde1: $K=\Sigma_x H \Sigma_d^{-1}$. Because we have $\Sigma_d^{-1}$ as the component here _and not_ $\Sigma_y^{-1}$ we need to pass in $D$.

docs/source/LinearRegression.py
```diff
@@ -170,3 +174,131 @@ def g(X):
...
+    assert H.shape == (m, p), "Shape of H_sparse must be (m, p)"
+
+    return H
+
+
+# %%
+# Learn Kalman gain
+X_prior = np.copy(X)
+Y = g(X_prior)
+D = Y + obs_sd * rng.standard_normal(size=Y.shape)
```
tommyod, line 268: Shouldn't the samples be divided by `sqrt(ensemble_size - 1)` here? We're drawing samples to approximate the covariance matrix.

Blunde1: No, I do not think so. We want to pass samples from $(x,d)$. I believe that is what is being done above.

tommyod: You are correct!

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+
+Note that $x|y$ retains Gaussianity as
+
+$$
+x | y \sim \mathcal{N}(\mu_x + K(y - \mu_y), \Sigma_{x} - K \Sigma_{yx}).
+$$
+
+For any distribution having a bijection between (the informative parts of) 
+$x$ and $y$, there exist a similar mapping, transporting a sample $(x,y)$ to $x|y$.
+The EnKF and ES variants follows from writing $y=d=Hx+\epsilon$ where
```
dafeda, line 56:
```suggestion
The EnKF and ES variants follow from writing $y=d=Hx+\epsilon$ where 
```

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+$$
+x | y \sim \mathcal{N}(\mu_x + K(y - \mu_y), \Sigma_{x} - K \Sigma_{yx}).
+$$
+
+For any distribution having a bijection between (the informative parts of) 
+$x$ and $y$, there exist a similar mapping, transporting a sample $(x,y)$ to $x|y$.
+The EnKF and ES variants follows from writing $y=d=Hx+\epsilon$ where 
+$\epsilon\sim \mathcal{N}(0,\Sigma_{\epsilon})$, thus $\Sigma_y=H\Sigma_xH^T+\Sigma_{\epsilon}$.
+Note that $y$ is common to use as $y=Hx$ but this is not the case in the preliminaries here.
+A point of confusion is that it is indeed _not_ the observation $d$ that is "perturbed" with noise,
```
dafeda, line 59: Having mostly read Evensen, I am used to seeing $y = g(x)$ followed by $d=y + \epsilon$ where $f(\epsilon) = \mathcal{N}(0, \Sigma_d)$.
Do I understand correctly that this aligns well with how you think about "perturbed observations"?

As for the term itself, it kinda makes sense to me from a practical point of view.
Users of ert specify an observation $d$ by providing a value and an error $\epsilon$.
To create $D\in\mathcal{1\times N}$, we sample $N$ times from $\mathcal{0, \epsilon}$ and add that to $d$.
In this sense, we "perturbe" $d$ to create $D$.

What would be a better name for $D\in\mathcal{1\times N}$?

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+but rather, $d$ has variance $H\Sigma_xH^T+\Sigma_{\epsilon}$ and an $\epsilon$ is then added to the response $Hx$ to correctly sample from this distribution of $d$.
+Because both $(d-d_i)$ and $d_i=Hx_i+\epsilon_i$ are additive, and $\epsilon$ is symmetric about 0, it does not matter if we add the noise to $d$ or to $Hx$ in the expression.
+If any of these points were different, it would be evident that we are not perturbing observations.
+Thus, it is incorrect and slightly confusing to say that we perturb the observations.
+It is correct to say that we sample from the distribution of $d$, accounting from both prior uncertainty _and_ observation uncertainty.
+
+Note that the EnKF and ES is intended to work (and indeed do) when $d=h(x)+\epsilon$
```
dafeda, line 66:
```suggestion
Note that the EnKF and ES are intended to work (and indeed do) when $d=h(x)+\epsilon$
```

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Why EnKF and ES works so well
+
+Let $(x,y)\sim P$ possibly non-Gaussian and assume we have a finite dataset to infer a model $Q$ from.
+Arguably, the most important aspects to encode in $Q$, without any other knowledge, are 
+the two first moments of $P$.
+Having access to the sample covariances of
+$\Sigma_{xy}$
+and $\Sigma_{y}^{-1}$
+after (maximum likelihood when $n>p$) estimation (minimizing empiricla KLD), the distribution $Q$ minimizing KLD towards $P$ is the Gaussian.
```
dafeda, line 130:
```suggestion
after (maximum likelihood when $n>p$) estimation (minimizing empirical KLD), the distribution $Q$ minimizing KLD towards $P$ is the Gaussian.
```

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+Thus, using a test dataset, we would like to evaluate different models $Q$, performing slightly different transports.
+
+
+## Objectives for Kalman-type ensembles
+
+When constraining $Q$ to only encode information on the two first moments, that are learned from data, the KLD
+approach yields a Gaussian distribution $Q$ even when $P$ is non-Gaussian.
+We call methods using the Kalman-gain in Gaussian transport from prior to posterior as "Kalman-type" ensemble based data assimilation methods.
+They implicitly seek to minimize information loss in this transport.
+Different Kalman-type modelling approaches essentially yields different estimates of the Kalman-gain, $K$.
```
dafeda, line 155:
```suggestion
Different Kalman-type modelling approaches essentially yield different estimates of the Kalman-gain, $K$.
```

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+- How to think about surrogate objectives?
+- What is the modelling setup and what information is known a-priori?
+- List common methods and discuss them in the context of the above.
+
+What is _not_ provided, is an evaluation-comparison between methods.
+Only how we _think_ they will behave, given what we know about statistics and our objective.
+From the contents, it should however be clear how a simulation study could be constructed to compare methods.
+
+
+## Gaussian conditional transport
```
dafeda, line 17: The term `Gaussian conditional transport` is new to me, and a search provides no direct hits.
Is this a known term in statistics?

docs/source/kalman_objective.md
```diff
@@ -0,0 +1,481 @@
...
+It could also be that $\Sigma_x$ is diagonal, or is a function of some $\theta$ in a smaller dimensional space.
+We would then require fewer degrees of freedom to fit $\Sigma_x(\theta)$.
+
+Note that if covariance estimates are singular (i.e. EnKF when $p>n$) then this objective cannot be used due to the log-determinant and covariance inverse.
+Furthermore, we often only have access to the final estimate of the Kalman-gain, not the full covariance structure.
+
+### Gaussian conditional marginal
+
+The joint KLD is a natural objective to evaluate model performance on.
+However, multiple (Kalman) ensemble methods only produce an estiamte $K$ either explicit or explicit,
```
dafeda, line 203:
```suggestion
However, multiple (Kalman) ensemble methods only produce an estimate $K$ either explicit or implicit,
```

Not sure I understand this sentence.
Do you mean that several different ensemble methods only produce an estimate of $K$ without specifying covariance structure?
What do you mean by explicit / implicit?

docs/source/kalman_objective.md
Blunde1: It is a shortening of something like "Transporting a joint-Gaussian sample to a (Gaussian-)conditional sample". Perhaps better to spell it out if it is confusing.

docs/source/kalman_objective.md
Blunde1: From a practical point of view, it does not matter. From a theoretical point of view of understanding why these methods work, it matters. (Showcasing how I think about these methods was the goal of the markdown). We need a joint sample from $(x,d)$, and then want to "map" or "transport" this, given an observation say $d^\ast$, to $x|d^\ast$.

To achieve this, a joint sample $(x_i,d_i)$ is needed first. Here $d=h(x)+\epsilon$ or $d=g(x)+\epsilon$ depending on which notation you use. The deterministic model produces $h(x_i)$, but to achieve a correct sample from $d$ we need $d_i=h(x_i)+\epsilon_i$ where $\epsilon_i$ are samples from the observation uncertainty. The innovation is then $d^\ast - d_i$. But, since this is additive, and the noise is additive in $d$, and $\epsilon$ has a symmetric distribution about 0, we may perturb $d^\ast$ instead and compute $d_i^\ast=d^\ast+\epsilon_i$ and see that $d^\ast - d_i$ has the same distribution as $d_i^\ast - g(x_i)$.

The latter term is used extensively, where $d_i^\ast$ are the perturbed observations. In practice this works fine and are distributionally equivalent, but theoretically it adds another layer of confusion to what is really occurring.

---

Note that I do not intend to change language of methods, because I did not intend this as some reference theory document (this is why there are no references, among other things). My goal was to showcase how I personally think. Distribute statistical tools and understanding of methodology given that I think correctly, or become aware of faults in my own logic.

The very particular goal is to create an understanding for LASSO++ to get future PRs with new methodology reviewed and merged. Do we have suggestions/an idea of requirements for such PRs?

Issue #200, opened 2023-12-17
Benchmark to make sure `groupby_indices` is necessary


dafeda:
The idea of `groupby_indices` was put together while trying to make Troll run in reasonable time.
Many things have changed since, and my benchmarking while working with Troll was not rigorous.
This trick introduces some complexity and we should make sure it's worth it.
I therefore think we should try running at least Drogon, and perhaps even Troll with and without it.

https://github.com/equinor/iterative_ensemble_smoother/blob/cf2c87270d31cbc1539e9d9cfb64f3167333233c/src/iterative_ensemble_smoother/experimental.py#L281-L295

- Trick should be most beneficial when parameters are highly correlated, i.e., a nice and smooth gaussian field
- Perhaps the trick provides more benefit in later iterations because updated parameters are more correlated? Speculative

dafeda:
## `With grouping` | `test_and_benchmark_adaptive_localization_with_fields` with 1 million parameters and 50 observations:

```
------------------------------------------------------------- benchmark: 1 tests ------------------------------------------------------------
Name (time in s)                                            Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------
test_and_benchmark_adaptive_localization_with_fields     1.3600  1.7665  1.4924  0.1622  1.4667  0.1782       1;0  0.6701       5           1
---------------------------------------------------------------------------------------------------------------------------------------------
```

## `Without grouping` | `test_and_benchmark_adaptive_localization_with_fields` with 1 million parameters and 50 observations:

--------------------------------------------------------------- benchmark: 1 tests --------------------------------------------------------------
Name (time in s)                                             Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------
test_and_benchmark_adaptive_localization_with_fields     26.2410  27.9223  26.9637  0.6318  26.8864  0.8130       2;0  0.0371       5           1
-------------------------------------------------------------------------------------------------------------------------------------------------

dafeda:
![image](https://github.com/equinor/iterative_ensemble_smoother/assets/45088507/0dd041d9-4274-4410-9165-14129870a905)

PR #205, opened 2024-01-02
Remove grouping trick

dafeda:
Resolves: #200

### DIFF
--- a/src/iterative_ensemble_smoother/experimental.py
+++ b/src/iterative_ensemble_smoother/experimental.py
@@ -4,7 +4,7 @@ features of iterative_ensemble_smoother
 """
 import numbers
 import warnings
-from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Union
+from typing import Callable, List, Optional, Tuple, Union
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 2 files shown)


dafeda:
Ran with and without trick on a gaussian field with 1 million parameters, 50 observations and 25 realizations. The sigma=10 controls the smoothness of the field.
It runs faster without the trick.

```
    p = 1 mill, m = 50, N = 25, sigma = 10
    With trick:    1.8s, 1.7s, 1.9s (Mean)
    Without trick: 1.3s, 1.2s, 1.3s (Mean)
```

image

I also ran Drogon with and without the trick and it runs much faster without the trick
My conclusion for now is that the trick is faster when the field is very smooth which it does not seem to be in practice (Drogon).
So, I suggest we remove it.

frode-aarstad (approved):

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -276,30 +239,21 @@ def correlation_threshold(ensemble_size: int) -> float:
...
-
-            # These parameters are not significantly correlated to any responses
-            if np.all(~unique_row):
-                continue
+        # Loop only over rows with significant correlations
+        for indices_of_row in np.where(significant_rows)[0]:
+            unique_row = significant_corr_XY[indices_of_row]
 
             # Get the parameters (X) that have identical significant responses (Y)
             X_subset = X[indices_of_row, :]
```
tommyod, line 252: Since `indices_of_row` is now an int, this variable `X_subset` is now a 1D array. I am surprised that an error is not raised or that the code does not fail?

Perhaps look at this quickly before merging. Also, perhaps rename 
```
indices_of_row -> parameter_num
unique_row -> correlated_responses
```
or something like that. No other comments.

tommyod (approved):

src/iterative_ensemble_smoother/experimental.py
dafeda: Have not looked into it deeply but I'm assuming broadcasting is helping us out in this: `X_subset + cov_XY_subset @ T`.
Thanks for the better variable names 👍

dafeda:
I want to run a final test using Drogon before merging, but we have some problems with `bleeding`.

PR #209, opened 2024-02-01
Upgrade ci versions

eivindjahren:


.github/workflows/precommit.yml
```diff
@@ -6,11 +6,11 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.8
+          python-version: "3.12"
```
pinkwah, line 13: docs.yml uses 3.11. This uses 3.12. Should both be the same?

pinkwah (approved):

.github/workflows/precommit.yml
eivindjahren: One of the dependencies of docs does not support 3.12 yet.

Issue #213, opened 2024-03-07
Make sure empirical cross covariance contains 32-bit values if X is a 32-bit float array


dafeda:
`Y` is always a 64-bit array which means that `cov` will also be 64-bit even if `X` is 32-bit.
This can cause memory issues.

https://github.com/equinor/iterative_ensemble_smoother/blob/a48841bd49405077c18e8513ef7ea26bc88e8807/src/iterative_ensemble_smoother/esmda_inversion.py#L80

PR #214, opened 2024-03-07
Optimize memory usage in empirical_cross_covariance calculation

dafeda:
Resolves: #213 

If X is a large matrix, it might be stored as a float32 array to save memory. However, if Y is of type float64,
the resulting cross-covariance matrix will be float64, potentially doubling the memory usage even if X is float32. To prevent unnecessary memory consumption,
we cast Y to the same data type as X before computing the dot product. This ensures that the output cross-covariance matrix uses memory efficiently while retaining the precision dictated by X's data type.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,6 +1,7 @@
 """
 iterative_ensemble_smoother documentation build configuration file.
 """
+
 import datetime
 import os
 import re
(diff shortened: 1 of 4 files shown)


tommyod (approved):
LGTM! Only comment is that you could generalize and convert to the "lowest" type: if either X or Y is float32, then convert the other. But I guess X being float32 is the realistic case.

PR #216, opened 2024-04-29
Add correlation callback to adaptive loc

dafeda:


tommyod (approved):
Looks good! Clever leveraging of existing tests as a sanity check too!

PR #222, opened 2025-01-02
Return all correlation pairs

dafeda:
Users have requested to see all correlation pairs, not just the ones that are significantly correlated.

Makes it possible to implement features like this one:

https://github.com/equinor/ert/pull/9426


tommyod (approved):
Looks good to me :)

PR #228, opened 2025-06-14
Run adaptive localization in parallel

dafeda:


src/iterative_ensemble_smoother/experimental.py
```diff
@@ -280,40 +322,32 @@ def progress_callback(x):
...
-
-            # Slice the covariance matrix
-            C_D_subset = (
-                self.C_D[correlated_responses]
-                if self.C_D.ndim == 1
-                else self.C_D[cov_YY_mask]
+        params_to_update = np.where(significant_rows)[0]
+
+        results = Parallel(n_jobs=n_jobs)(
+            delayed(_update_single_parameter)(
```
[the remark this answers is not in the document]
dafeda: Here's what I get when I raise a ValueError and then run the test `test_parallelization_runtime_comparison`.

```console
    def _update_single_parameter(
        param_num: int,
        correlated_responses_mask: npt.NDArray[np.bool_],
        cov_XY_row: npt.NDArray[np.double],
        Y: npt.NDArray[np.double],
        D: npt.NDArray[np.double],
        C_D: npt.NDArray[np.double],
        cov_YY: npt.NDArray[np.double],
        alpha: float,
        compute_cross_covariance_multiplier: Callable,
    ) -> Tuple[int, npt.NDArray[np.double]]:
        """
        Worker function to compute the update for a single parameter.
        """
    
>       raise ValueError("Testing joblib exception handling...")
E       ValueError: Testing joblib exception handling...

src/iterative_ensemble_smoother/experimental.py:39: ValueError
----------------------------- Captured stdout call -----------------------------

--- Running Serial Benchmark (n_jobs=1) ---
=========================== short test summary info ============================
FAILED tests/test_experimental.py::TestAdaptiveESMDA::test_parallelization_runtime_comparison
============================== 1 failed in 0.38s ===============================
Finished running tests!
```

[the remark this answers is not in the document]
xjules: Nice! So (if I understand it correctly), we will get only one exception and not altered; so then it is great :)

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -177,6 +139,47 @@ def _corr_to_cov_inplace(
         corr_XY *= stds_Y[np.newaxis, :]
         return corr_XY
 
+    @classmethod
```
JHolba, line 142: you could use @staticmethod instead of @classmethod since you do not use cls for anything

dafeda: Good point, thanks. I've pushed a fix.

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -138,6 +139,46 @@ def _corr_to_cov_inplace(
         corr_XY *= stds_Y[np.newaxis, :]
         return corr_XY
 
+    @staticmethod
+    def _update_single_parameter(
```
xjules, line 143: `_update_single_parameter_group`?

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -138,6 +139,46 @@ def _corr_to_cov_inplace(
...
+        param_num: int,
```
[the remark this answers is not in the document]
xjules: It seems that `Parallel(n_jobs=n_jobs, return_as="list")` (default setting) preserves the order of the outputs.

tests/test_experimental.py
```diff
@@ -431,6 +431,92 @@ def test_that_cov_YY_can_be_computed_outside_of_assimilate(
...
+        # --- 1. Serial Execution ---
+        print("\n--- Running Serial Benchmark (n_jobs=1) ---")
+        start_serial = time.perf_counter()
+        X_serial = smoother.assimilate(
+            X=X_serial,
+            Y=Y,
+            D=D,
+            alpha=alpha,
+            correlation_threshold=0,  # Force the code to update every single parameter.
+            n_jobs=1,
```
xjules, line 487: Not sure how long this takes. But can you do at least two n_jobs values: `-1` and `1` with:
1) assert -1 should be faster
2) results with -1 and 1 are the same

tests/test_experimental.py
```diff
@@ -431,6 +431,92 @@ def test_that_cov_YY_can_be_computed_outside_of_assimilate(
...
+        D = smoother.perturb_observations(ensemble_size=Y.shape[1], alpha=alpha)
+
+        # Make separate copies for each run to ensure a fair start
+        X_serial = np.copy(X)
+        X_parallel = np.copy(X)
+
+        # --- 1. Serial Execution ---
+        print("\n--- Running Serial Benchmark (n_jobs=1) ---")
+        start_serial = time.perf_counter()
+        X_serial = smoother.assimilate(
```
xjules, line 481: Maybe also another test where we mock `def _update_single_parameter(` with `raise ValueError` and see that it raises?

tests/test_experimental.py
xjules: Missed the bottom part! :)

PR #233, opened 2025-10-20
Remove overwrite bug in SVD

### LINKED ISSUE #232: Physical memory layout of S affects transition matrix
How to reproduce:

Run `test_update_snapshot` and breakpoint here:
```
T = smoother_es.compute_transition_matrix(
                Y=S, alpha=1.0, truncation=truncation
            )
```

Try passing in S as fortran: `np.asfortranarray(S)`, then as C-contiguous (current default), see different results. The physical memory layout should not affect numeric results ideally. Origin of the mismatch is somewhere inside IES, likely `inversion_subspace`, perhaps this part:
```
    U_r_w_inv = U_r / w_r
    if C_D.ndim == 1:
        X1 = (N_e - 1) * np.linalg.multi_dot([U_r_w_inv.T * C_D * alpha, U_r_w_inv])
    else:
        X1 = (N_e - 1) * np.linalg.multi_dot([U_r_w_inv.T, alpha * C_D, U_r_w_inv])
```
linalg.multi_dot does state it requires a c-contiguous array, but does not error there.

We should be more explicit about this somehow, as changes in snapshots wrt changes in order are hard to track down. (polars .to_numpy() defaults to F-order)

tommyod:
Closes #232

### DIFF
--- a/docs/source/Polynomial.py
+++ b/docs/source/Polynomial.py
@@ -165,15 +165,11 @@ for i in range(n_ies_iter):
     X_IES_ert = smoother_ies.sies_iteration(Y_IES_ert, step_length=step_length)
 
     _coeff_a, _coeff_b, _coeff_c = X_IES_ert
+    x = [np.arange(max(x_observations) + 1)] * ensemble_size
 
-    _fwd_runs = p_map(
-        poly,
-        _coeff_a,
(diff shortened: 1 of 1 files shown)


andreas-el (approved):
🏖️

Issue #227, opened 2025-06-06
Implement distance based localization algorithm


dafeda:
Algorithm 3 of emerick16a - Analysis of the Performance of Ensemble-based Assimilation of Production and Seismic Data

- Test algorithm using synthetic test cases with increasing complexity. (See tests of adaptive localization for inspiration)

## API design

PR #240, opened 2025-12-04
Add distance based esmda algorithm

oddvarlia:
The distance-based algorithm published by Emerick 2016 is implemented in class DistanceESMDA.
Linked to #227

### DIFF
--- a/src/iterative_ensemble_smoother/esmda.py
+++ b/src/iterative_ensemble_smoother/esmda.py
@@ -108,7 +108,8 @@ class BaseESMDA(ABC):
         Returns
         -------
         D : np.ndarray
-            Each column consists of perturbed observations, scaled by alpha.
+            Each column consists of perturbed observations,
+            observation std is scaled by sqrt(alpha).
 
         """
         # Draw samples from zero-centered multivariate normal with cov=alpha * C_D,
@@ -222,6 +223,7 @@ class ESMDA(BaseESMDA):
         *,
         overwrite: bool = False,
         truncation: float = 1.0,
+        D: npt.NDArray[np.double] = None,
     ) -> npt.NDArray[np.double]:
         """Assimilate data and return an updated ensemble X_posterior.
 
@@ -279,9 +281,11 @@ class ESMDA(BaseESMDA):
         # if C_D = L L.T by the cholesky factorization, then drawing y from
         # a zero cented normal means that y := L @ z, where z ~ norm(0, 1)
         # Therefore, scaling C_D by alpha is equivalent to scaling L with sqrt(alpha)
-        D = self.perturb_observations(
-            ensemble_size=num_ensemble, alpha=self.alpha[self.iteration]
-        )
+        if D is None:
+            D = self.perturb_observations(
+                ensemble_size=num_ensemble, alpha=self.alpha[self.iteration]
+            )
+
         assert D.shape == (num_outputs, num_ensemble)
 
         # Line 2 (c) in the description of ES-MDA in the 2013 Emerick paper
--- a/src/iterative_ensemble_smoother/experimental.py
+++ b/src/iterative_ensemble_smoother/experimental.py
@@ -16,6 +16,7 @@ from iterative_ensemble_smoother import ESMDA
 from iterative_ensemble_smoother.esmda import BaseESMDA
 from iterative_ensemble_smoother.esmda_inversion import (
     empirical_cross_covariance,
+    singular_values_to_keep,
 )
 
 T = TypeVar("T")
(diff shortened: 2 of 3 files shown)


src/iterative_ensemble_smoother/esmda.py
```diff
@@ -279,9 +281,11 @@ def assimilate(
         # if C_D = L L.T by the cholesky factorization, then drawing y from
         # a zero cented normal means that y := L @ z, where z ~ norm(0, 1)
         # Therefore, scaling C_D by alpha is equivalent to scaling L with sqrt(alpha)
-        D = self.perturb_observations(
-            ensemble_size=num_ensemble, alpha=self.alpha[self.iteration]
-        )
+        if D is None:
```
xjules: I guess this does not need to be there?

oddvarlia: This test on `D is None` is there for ESMDA to itself calculate `D `instead of having` D` as input. For backward compatibility `D `is not assigned but calculated inside ESMDA class.

xjules: Thanks for clarification. I think then we can keep it then - to have the possibility to get observations from outside 👍

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
...
+
+    rng = np.random.default_rng(seed)
+    X_prior = rng.normal(loc=0.0, scale=0.5, size=(N_m, N_e))
+
+    # Predict observations `Y` using the identity model `g(x) = x`
+    # We only observe the state at `j_obs`.
+    Y = X_prior[[j_obs], :]
+
+    # Using a simple Gaussian decay for rho
+    localization_radius = 10.0
```
xjules: This tapering function, I was wondering, currently the implementation of this function is in ert basically. Maybe it should be present also in `iterative_ensemble_smoother`. What do you think?

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
         )
 
 
+@pytest.mark.parametrize("seed", list(range(9)))
+def test_distance_based_localization_on_1D_case_with_single_observation(seed):
+    N_m = 100  # Number of model parameters (grid points)
+    N_e = 50  # Ensemble size
+    j_obs = 50  # Index of the single observation
+
+    alpha_i = 1  # TODO: Implement multiple iterations
```
xjules: I'd remove the comment and instead have a separate test that multiple iterations.

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
...
+
+    # --- Overall Quality Assertion (MSE) ---
+    # Assert that the localized result is a more plausible estimate
+    mse_localized = np.mean((posterior_mean - true_parameters) ** 2)
+    mse_non_localized = np.mean((posterior_mean_global - true_parameters) ** 2)
+    assert mse_localized < mse_non_localized
+
+
+@pytest.mark.parametrize("seed", list(range(9)))
+def test_distance_based_localization_on_3D_case(seed):
```
xjules: There is this sufffix everywhere `_with_single_observation` witch either should be dropped and provided in all tests?

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
...
+
+    # --- 5. Assertions ---
+    # Find parameters far from the observation where localization weight is near zero
+    zero_weight_indices = np.where(rho < 1e-6)[0]
+    assert len(zero_weight_indices) > 0, (
+        "No distant points found for testing localization."
+    )
+
+    # --- 5a. Assertions for Localized Smoother ---
+    prior_mean = np.mean(X_prior, axis=1)
```
xjules: This assertions are the same for all tests. I'm wondering if those can be taken to a helper function and thus having just a different setup?

tests/test_experimental.py
oddvarlia: The tapering function implemented so far for ert application is the Gaspari-Cohn function (we skipped Gaussian decay or exponential decay in that code).  I don't think we need details about how RHO is calculated in implementation of DistanceESMDA class. The class is implemented for a general RHO. The argument I have for modification of input to DistanceESMDA is to have options to utilize the fact that we assume 2D distance-calculations when creating RHO. We copy one grid layer of ertbox with RHO parameters for all observations into all the other layers in a batch of complete layers of field parameters. A possible memory saving optimization can be to given the assimilate function RHO for one layer only and specify 3D grid dimensions for the field parameters ( or batch of field parameters). Then it is possible to do the SCHUR product (element by element multiplication) of RHO with K matrix layer by layer of field parameters and we can avoid having to assign values to RHO for all nx * ny * nlayer_in_batch * nobs but can limit it to one layer instead and let numpy take care of the loop over all layers in a batch of layers. Some memory can be saved, but still we have the K matrix for all parameters in the batch.

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
...
+    posterior_variance = np.var(X_posterior, axis=1)
+
+    # The variance of the ensemble should decrease where the update was applied
+    assert posterior_variance[j_obs] < prior_variance[j_obs]
+    # The variance of the ensemble should NOT change where there was no update
+    assert np.allclose(
+        posterior_variance[zero_weight_indices], prior_variance[zero_weight_indices]
+    )
+
+    esmda = ESMDA(
```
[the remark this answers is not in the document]
oddvarlia: I assume this is not necessary here, but I think the tests is taken from a jupyter notebook comparing the methods. Should I remove the code related to ESMDA test and the comparison tests?

tests/test_experimental.py
oddvarlia: A test with multiple data-assimilation without any forward model, just repeated update with alpha[i] defined according to Emerick's paper  should be easy to create, but what should I compare the result with. Maybe I could find a case where I know the ensemble mean and ensemble std.dev like e.g what we get by calculating predictions using geostatistical kriging with known variogram

src/iterative_ensemble_smoother/experimental.py
```diff
@@ -451,6 +452,290 @@ def ensemble_smoother_update_step_row_scaling(
     return X_with_row_scaling
 
 
+class DistanceESMDA(ESMDA):
```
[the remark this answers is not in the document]
oddvarlia: OK. and separate test file for each. But how long should a method be located under experimental before it is given a more permanent filename? Is this related to TRL level of the method and implementation?

tests/test_experimental.py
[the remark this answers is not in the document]
xjules: I need to  think about that, but maybe it is fine to keep it there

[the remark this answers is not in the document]
oddvarlia: OK

xjules:
> Should we have tests where the field parameters are not only random uncorrelated parameters, but have spatial correlations?

Not sure if I understand what you mean? Maybe we can have a teams session on this.

tests/test_experimental.py
oddvarlia: I have added a test looping over alpha for alpha = 7, 3.5, 1.75 for 1D case an used the assert tests as they are. I have made separate function for the asserts. Is this sufficient?

tests/test_experimental.py
oddvarlia: Fixed with a separate function doing the asserts

src/iterative_ensemble_smoother/experimental.py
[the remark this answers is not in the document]
oddvarlia: I will wait with this and define a new PR after the current one is merged.

tests/test_experimental.py
```diff
@@ -630,6 +631,290 @@ def multiply(self, X, K):
...
+    # Find parameters far from the observation where localization weight is near zero
+    zero_weight_indices = np.where(rho < 1e-5)[0]
+    assert len(zero_weight_indices) > 0
+
+    # Assert that parameters in the zero-weight region have not been updated
+    assert np.allclose(
+        X_posterior[zero_weight_indices, :], X_prior[zero_weight_indices, :], rtol=1e-2
+    )
+    # Assert that parameters near the observation *have* been updated
+    assert not np.allclose(X_posterior[j_obs, :], X_prior[j_obs, :])
```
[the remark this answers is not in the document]
oddvarlia: Since I have made separate function for asserts, this comment is replaced with another one now, but I think what was meant that the value of field parameter in observation point is closer to the observation value
 for posterior than for prior, so nothing to do with spatial distance, only distance in value of the field versus observation.

tests/test_experimental.py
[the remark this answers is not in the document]
oddvarlia: I suggest we keep it since this is a form for automatic tests comparing the two methods and gives expected difference between them.

tests/test_experimental.py
```diff
@@ -630,6 +632,330 @@ def multiply(self, X, K):
...
+    # Calculate Mean Squared Error (MSE) against the true parameters
+    mse_localized = np.mean((posterior_mean - true_parameters) ** 2)
+    mse_non_localized = np.mean((posterior_mean_global - true_parameters) ** 2)
+
+    # --- Overall Quality Assertion (MSE) ---
+    # Assert that the localized result is a more plausible estimate
+    assert mse_localized < mse_non_localized
+
+
+def calculate_rho_1d(N_m: int, obs_index: int, localization_radius: float):
```
xjules: nice! Since you already doing the type annotations, could you do the return type (for all rho calculations)?

xjules:
> > Should we have tests where the field parameters are not only random uncorrelated parameters, but have spatial correlations?
> 
> Not sure if I understand what you mean? Maybe we can have a teams session on this.

Resolved in conversion. This aspect, ie. having correlation among parameters, will be handled in a separate issue.

PR #247, opened 2026-01-05
Add test of distance-based localization with spatial correlated 1D fields

oddvarlia:
Purpose of the added test is to verify that distance-based ESMDA converges for posterior mean to something close to a known theoretical value for a simple test where a theoretical limit is known. In this test case simple kriging estimate of the posterior mean
can be used to calculate the theoretical posterior mean. To compare with ordinary ESMDA, also ESMDA  posterior estimate of the mean value is calculated, and for ordinary ESMDA in the limit of infinitely many realizations, the posterior mean will converge to the simple kriging estimate. For Distance-based localization with ESMDA the posterior mean will not converge to the simple kriging estimate due to the finite localization range, but will be close to it (depending on the ratio between the spatial correlation length of the prior field and the localization range. The Distance based localization estimate will however be closer to the simple kriging estimate than ESMDA for small number of realizations (100 to 1000 realizations). The test is fast enough that up to 40000 realizations is used. There is a practical upper limit of number of realizations in this test, but 40000 realizations worked.


oddvarlia:
Note: PR 246 should be merged first since PR 247 builds on PR 246

PR #252, opened 2026-01-14
Add new tests to DistanceESMDA including check for convergence

oddvarlia:
New tests to check that ESMDA and DistanceESMDA give same result when RHO matrix consists of only 1.
New test to check convergence (when increasing number of realizations) to known theoretical result (simple kriging estimate of ensemble average) for special case.
 
Updated DistanceESMDA such that internal X3 matrix is automatically created if not already created.
The use of DistanceESMDA can now be used in two different ways:
Without need for batch split-up of field parameters:
- Initialize DistanceESMDA object
- call assimilate function to update
Alternative is:
- Initialize DistanceESMDA object
- Loop over field parameter batches:
  - call assimilate_batch function to update each batch of field parameters
  - collect the update into one array

Both will do the same except that for the latter one, the update steps in the algorithm related to observations only will be done only the first time the assimilate_batch is called since in a loop over batches of field parameters, the only difference in each call to assimilate_batch is to do the steps related to field parameters and it is not necessary to do the steps with the observations more than once. When using the assimilate function instead, both the algorithm steps related to observations and field parameters are done each time (which is a bit waste of time with large observation data set like when using seismic observations).

To test memory use, a memory_usage_decorator function is added which can turn on/off a check and print of memory usage before and after calling a function.

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -704,10 +704,10 @@ def assert_tests_for_distance_based_localization(
 
 
 def calculate_rho_1d(
-    N_m: int, obs_index: int, localization_radius: float
+    N_m: int, obs_index: int, localization_radius: float, xinc: float = 1.0
 ) -> npt.NDArray[np.float64]:
-    model_grid = np.arange(N_m)
-    distances = np.abs(model_grid - obs_index)
+    model_grid = np.arange(N_m) * xinc
+    distances = np.abs(model_grid - obs_index * xinc)
     # Ordinary definition of range,
     # rho is approximately 0.05 at localization_radius
     return np.exp(-3.0 * (distances / localization_radius) ** 2).reshape(-1, 1)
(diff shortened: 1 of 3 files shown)


tests/test_experimental.py
```diff
@@ -955,6 +1032,385 @@ def test_distance_based_on_1D_case_multiple_data_assimilation(seed):
         )
 
 
+@pytest.mark.parametrize("seed", list(range(9)))
+def test_distance_based_on_1D_case_multiple_data_assimilation_alternative(seed):
```
xjules: `test_distance_based_on_1D_case_multiple_data_assimilation_with_batching`

tests/test_experimental.py
```diff
@@ -955,6 +1032,385 @@ def test_distance_based_on_1D_case_multiple_data_assimilation(seed):
...
+    j_obs = 50  # Index of the single observation
+    nalpha = 10  # Number of different sets of alpha values
+    number_of_alpha_values = np.arange(1, nalpha, 1, dtype=np.int32)
+    for m in number_of_alpha_values:
+        alpha_vector = generate_alpha_vector(m)
+        inverse_alpha_vector = 1.0 / alpha_vector
+
+        # Consistency requirement for alpha
+        # Sum of 1/alpha_vector[i] must be 1
+        assert abs(inverse_alpha_vector.sum() - 1.0) < 1e-7
```
xjules: this is not longer necessary. If you want you can include the assert directly into the `generate_alpha_vector` function

tests/test_experimental.py
```diff
@@ -955,6 +1032,385 @@ def test_distance_based_on_1D_case_multiple_data_assimilation(seed):
...
+    C_D = np.array([obs_error_var] * nobs)
+
+    # Draw prior ensemble with specified spatial correlation function
+    rng = np.random.default_rng(seed)
+    X_prior, field_cov_matrix = draw_1D_field(
+        mean, stdev, xinc, corr_func_name, corr_range, nparam, nreal, rng
+    )
+
+    # Calculate theoretical posterior mean
+    # assuming rior mean equal 0.
```
xjules: typo: rior -> prioir

xjules (changes requested):
Very good tests @oddvarlia !
I just had a smaller comments. I'll do another round once resolved.

tests/test_experimental.py
oddvarlia: Modified the name, OK

tests/test_experimental.py
oddvarlia: fixed

tests/test_experimental.py
```diff
@@ -955,6 +1032,385 @@ def test_distance_based_on_1D_case_multiple_data_assimilation(seed):
...
+    # of the effect of the observations with distance. (Reduced value
+    # of Kalman Gain matrix elements)
+
+    # Prior field mean and stdev
+    mean = 0.0
+    stdev = 1.0
+
+    # If this flag is set to False, rho matrix will be 1 for all elements
+    # and the result should be identical to ordinary ESMDA.
+    use_localization = False
```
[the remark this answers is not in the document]
oddvarlia: OK, I will do that.

PR #255, opened 2026-01-22
Replace print statements with logging

### LINKED ISSUE #253: Replace print statements with logging instead
Since we rely on print statements debugging now, this can become hard to manage once we start to used both enif and distanceESMDA and thus we should replace print statements with proper logging. 
Whether to bring logs into clouds remains to be discussed.

frode-aarstad:
Resolves #253

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -122,10 +122,14 @@ select = [
     "RSE",
     "SIM",
     "SLOT",
+    "T",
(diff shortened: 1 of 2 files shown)


berland:
Try adding `"T"` to the ruff ruleset list in `pyproject.toml`. It will refuse future `print` statements to the codebase.

PR #265, opened 2026-02-04
Remove no update

### LINKED ISSUE #262: Remove no_update option from DistanceESMDA
This is a remnant from testing and we don't want to support it. It would require to modify some tests though.
See:
https://github.com/equinor/iterative_ensemble_smoother/blob/8eac9e708214c30167b75454884b2061b1457567/src/iterative_ensemble_smoother/experimental.py#L982

frode-aarstad:
Resolves #262


xjules (approved):
Nice @frode-aarstad ! Remember to squash the commits.

PR #266, opened 2026-02-04
Fix DistanceESMDA.assimilate() to use per-iteration alpha values

dafeda:
DistanceESMDA.assimilate() was incorrectly using self.alpha (array) instead of self.alpha[self.iteration] (scalar) when computing the update and perturbing observations. This caused a broadcasting error when calling assimilate() multiple times with alpha > 1.

The fix:
- Use self.alpha[self.iteration] in the R matrix computation
- Use self.alpha[self.iteration] in perturb_observations()
- Increment self.iteration after each assimilation step

This makes DistanceESMDA consistent with the base ESMDA class behavior.

Added test_that_distance_esmda_alpha_as_integer_and_array_returns_same_result to verify the fix and prevent regression.

### DIFF
--- a/tests/test_experimental.py
+++ b/tests/test_experimental.py
@@ -1967,6 +1967,72 @@ def test_update_params_3D(
     snapshot.assert_match(str(X_post_stdev_3D) + "\n", "X_post_stdev_3D.txt")
 
 
+@pytest.mark.parametrize(
+    "num_ensemble",
+    [10, 100, 1000],
+)
+def test_that_distance_esmda_alpha_as_integer_and_array_returns_same_result(
+    num_ensemble,
+):
+    """Test that DistanceESMDA gives same results when alpha is an integer
+    vs when it's an array of ones with the same length.
+
+    This test mirrors test_that_alpha_as_integer_and_array_returns_same_result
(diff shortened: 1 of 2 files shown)


tests/test_experimental.py
```diff
@@ -1967,6 +1967,72 @@ def test_update_params_3D(
...
+    observations = G(X_true)
+
+    # Create a simple localization matrix rho (num_inputs x num_outputs)
+    # Use a simple distance-based decay
+    localization_radius = 5.0
+    rho = np.zeros((num_inputs, num_outputs))
+    for i in range(num_inputs):
+        for j in range(num_outputs):
+            distance = abs(i - j)
+            rho[i, j] = np.exp(-0.5 * (distance / localization_radius) ** 2)
```
oddvarlia: Usually the range is defined such that the correlation at one correlation range is close to 0, in practise, exp(-3.0*d) for exponential correlation function and  exp(-3.0*np.power(d, nu)) for general exponential correlation function where 1< nu < 2 and d is normalized distance. (But this is not important for the test if  aware of the definition of the correlation length when using exp(-0.5*d)

PR #268, opened 2026-02-05
Do not multiply with identity matrix

tommyod:
Creating a (obs, obs) identity matrix and multiplying with it a lot of unneccessary work, both in terms of time and memory.


dafeda (approved):

tommyod:
This seems secondary to the code duplication issue in the class. Closing it.

PR #278, opened 2026-02-09
Remove function 'generate_alpha_vector'

tommyod:
There is already a very similar function for this purpose, might as well use it.


tommyod:
Closing this, as the class will likely be superseded by LocalizedESMDA anyway.

PR #279, opened 2026-02-09
Update test to create ESMDA instances outside of loop

tommyod:
The current way this test was written does not conform to how ESMDA works. The idea is to instantiate the class once, then use it to assimilate several times. The class is responsible for computing factorizations once (at instantiation time), normalizing alpha values once, and incrementing the `iteration` counter, which fetches the correct alpha value on each iteration.

As this code is written, the ESMDA instance is created again per loop, with a single alpha value. This will normalize that value in each loop.


dafeda:
Since we have the separate method `assimilate_batch`, I think the changes proposed by Tommy here make sense.
That is, we still want to use the `assimilate` method the same way as we do in the standard `ESMDA`.
Does this make sense to you Oddvar?

PR #282, opened 2026-02-11
Localized ESMDA

tommyod:
- An API that decouples grid management from the smoother.
- Natural extension of ESMDA.
- Tests against current ESMDA implementation.
- Property tests on different inversion schemes, batch sizes, etc.

### DIFF
--- a/tests/test_esmda_localized.py
+++ b/tests/test_esmda_localized.py
@@ -2,6 +2,7 @@ import numpy as np
 import pytest
 
 from iterative_ensemble_smoother.esmda import ESMDA
+from iterative_ensemble_smoother.esmda_inversion import normalize_alpha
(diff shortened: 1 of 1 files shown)


src/iterative_ensemble_smoother/esmda_localized.py
```diff
@@ -0,0 +1,545 @@
...
+    def __init__(
+        self,
+        *,
+        covariance: npt.NDArray[np.double],
+        observations: npt.NDArray[np.double],
+        alpha: Union[int, npt.NDArray[np.double]] = 5,
+        seed: Union[np.random._generator.Generator, int, None] = None,
+        inversion: str = "exact",
+    ) -> None:
+        """Initialize the instance."""
```
dafeda:
```suggestion
```

PR #290, opened 2026-02-16
FIX: minor typo in SIES __init__ docstring

antoinecollet5:
For inversion parameter in `SIES.__init__`, the default is `subspace_exact` and not `subspace`.


dafeda (approved):
Thank you 👍

PR #291, opened 2026-02-18
Write about exact and subspace inversion

tommyod:


src/iterative_ensemble_smoother/esmda_inversion.py
```diff
@@ -1,3 +1,254 @@
...
+has a structure like A := (F F.T + C_D), where C_D is positive definite
+and F is tall and thin (many more rows than columns). Here is an example:
+
+>>> F = rng.normal(size=(4, 2))
+>>> G = rng.normal(size=(4, 4))
+>>> C_D = G.T @ G
+>>> A = (F @ F.T + C_D)
+
+Subspace inversion means using the Woodbury identity and an SVD when inverting A.
+The version of Woodbury that we'll use is the following:
```
dafeda: Is it worth being explicit about why we are using Woodbury here, i.e., inverting a smaller matrix?

src/iterative_ensemble_smoother/esmda_inversion.py
```diff
@@ -1,3 +1,254 @@
...
+       [ 0. ,  1. ,  0.1,  0.1],
+       [-0.3,  0.1,  0.4, -0.2],
+       [-0.1,  0.1, -0.2,  0.4]])
+
+At this point, one approach would be to use Woodbury identity directly, which
+avoids inverting a large matrix and instead inverts a much smaller matrix.
+If G has shape (m, n), with m >> n, then this has cost O(n^3) for the inversion,
+plus a cost of O(nm^2) to form the full product I - G (G.T G + I)^-1 G.T.
+
+The more common alternative, which is not less work computationally, is to
```
dafeda: Should we instead say that it's more costly computationally? Is that important here?

PR #298, opened 2026-02-20
Mask missing parameters in some realizations

tommyod:
Arguably this could be the callers responsibility. But we can put it into the class. I think it belongs there.

See also:
- https://github.com/equinor/ert/issues/12122
- https://github.com/equinor/ert/pull/12843


oddvarlia (commented):
Se my comments. I am not sure this is correct, but I can misunderstand something here.

PR #302, opened 2026-02-26
Add simulate_and_update_1d_field.py

oddvarlia:
Script to manually test 1D field upate with the various implementation of ESMDA with and without distance-based localization.


dafeda:
Closing as this was not meant to be merged but was pushed for discussions.

PR #335, opened 2026-03-26
Parallelize adaptive ESMDA update loops with joblib

dafeda:
Extract the per-parameter update logic in AdaptiveESMDA and TaperedAdaptiveESMDA into top-level worker functions (_update_group, _update_param) and dispatch them via joblib.Parallel.

Add an n_jobs parameter to both assimilate_batch methods (default -1, i.e. all cores).

Remove groupby_rows from utils.py — benchmarking showed no speedup from grouping parameters that share the same correlation pattern.

Drogon:

image

This is about twice as fast as the current komodo stable which is very nice.

### DIFF
--- a/src/iterative_ensemble_smoother/esmda_adaptive.py
+++ b/src/iterative_ensemble_smoother/esmda_adaptive.py
@@ -31,6 +31,7 @@ from typing import Callable, Union
 import numpy as np
 import numpy.typing as npt
 import scipy as sp
+from joblib import Parallel, delayed
 
 from iterative_ensemble_smoother.esmda import BaseESMDA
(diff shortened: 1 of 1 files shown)


tommyod (approved):
Approved. One concern is that groupby might help a lot on the problems where it does help? My intuition is poor. A problem needs quite a bit of structure for it to help, I guess.

dafeda:
> Approved. One concern is that groupby might help a lot on the problems where it does help? My intuition is poor. A problem needs quite a bit of structure for it to help, I guess.

It should help more in problems with structure and not that many observations.
My intuition here is poor as well.
Let's keep it for now, as it does not seem to hurt performance on Drogon and might help in other cases.

PR #353, opened 2026-05-18
Specify lower bound version for nbconvert to 7.17.1

xjules:
Resolves https://github.com/equinor/scout/issues/1509

`uv lock --upgrade-package nbconvert`

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -14,7 +14,7 @@ resolution-markers = [
 ]
 
 [options]
-exclude-newer = "0001-01-01T00:00:00Z" # This has no effect and is included for backwards compatibility when using relative exclude-newer values.
+exclude-newer = "2026-05-11T07:44:21.038992276Z"
 exclude-newer-span = "P7D"
 
 [options.exclude-newer-package]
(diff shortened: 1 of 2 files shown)


uv.lock
```diff
@@ -2993,14 +2988,14 @@ wheels = [
 
 [[package]]
 name = "tinycss2"
-version = "1.5.1"
+version = "1.4.0"
```
andreas-el, line 2991: Very curious what the reason for this is. We used the newest (6mnth old) version, but now we step back an additional year?

andreas-el: Probably some relation to;
    { name = "bleach", extra = ["css"] },

xjules: not sure tbh, this was done by automatically by calling `uv lock --upgrade-package nbconvert`.

andreas-el: I think this is the reason:
```
tinycss2: 1.4.0 # pinned by bleach==6.3.0
```

andreas-el (approved):
💯

